PR #470, opened 2015-01-21
AGPUSH-1150 : optional system properties for APNs gateway destination

sebastienblanc:
Let's start with APNs for now.
@TadeasKriz @matzew can you review ?

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
@@ -196,17 +196,52 @@ public class APNsPushNotificationSender implements PushNotificationSender {
                 return null;
             }
 
-            // pick the destination:
-            if (iOSVariant.isProduction()) {
-                builder.withProductionDestination();
-            } else {
-                builder.withSandboxDestination();
-            }
+            configureDestinations(iOSVariant, builder);
+
 
             // create the service
             return builder.build();
         }
         // null if, why ever, there was no cert/passphrase
         return null;
+    }
+
+    /**
+     * Configure the Gateway to the Apns servers.
+     * Default gateway and port can be override with respectively :
+     *  - aerogear.apns.push.host
+     *  - aerogear.apns.push.port
+     *
+     * Feedback gateway and port can be override with  respectively :
+     *  - aerogear.apns.feedback.host
+     *  - aerogear.apns.feedback.port
+     * @param iOSVariant
+     * @param builder
+     */
+    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
+        // pick the destination:
+        if (iOSVariant.isProduction()) {
+            builder.withProductionDestination();
+        } else {
+            builder.withSandboxDestination();
+        }
+
+        //Is the gateway host&port provided by a system property ?
+        if(System.getProperty("aerogear.apns.push.host")!= null){
+            int port = Utilities.SANDBOX_GATEWAY_PORT;
+            if(System.getProperty("aerogear.apns.push.port")!= null) {
+                port = Integer.parseInt(System.getProperty("aerogear.apns.push.port"));
+            }
+            builder.withGatewayDestination(System.getProperty("aerogear.apns.push.host"),port);
+        }
+
+        //Is the feedback gateway provided by a system property ?
+        if(System.getProperty("aerogear.apns.feedback.host")!= null){
+            int port = Utilities.SANDBOX_FEEDBACK_PORT;
+            if(System.getProperty("aerogear.apns.feedback.port")!= null) {
+                port = Integer.parseInt(System.getProperty("aerogear.apns.feedback.port"));
+            }
+            builder.withFeedbackDestination(System.getProperty("aerogear.apns.feedback.host"),port);
+        }
     }
 }
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
@@ -43,6 +43,10 @@ import java.util.Set;
 @SenderType(iOSVariant.class)
 public class APNsPushNotificationSender implements PushNotificationSender {
 
+    private static final String AEROGEAR_APNS_PUSH_HOST = "aerogear.apns.push.host";
+    private static final String AEROGEAR_APNS_PUSH_PORT = "aerogear.apns.push.port";
+    private static final String AEROGEAR_APNS_FEEDBACK_HOST = "aerogear.apns.feedback.host";
+    private static final String AEROGEAR_APNS_FEEDBACK_PORT = "aerogear.apns.feedback.port";
     private final AeroGearLogger logger = AeroGearLogger.getInstance(APNsPushNotificationSender.class);
 
     @Inject
@@ -227,21 +231,21 @@ public class APNsPushNotificationSender implements PushNotificationSender {
         }
 
         //Is the gateway host&port provided by a system property ?
-        if(System.getProperty("aerogear.apns.push.host")!= null){
+        if(System.getProperty(AEROGEAR_APNS_PUSH_HOST)!= null){
             int port = Utilities.SANDBOX_GATEWAY_PORT;
-            if(System.getProperty("aerogear.apns.push.port")!= null) {
-                port = Integer.parseInt(System.getProperty("aerogear.apns.push.port"));
+            if(System.getProperty(AEROGEAR_APNS_PUSH_PORT)!= null) {
+                port = Integer.parseInt(System.getProperty(AEROGEAR_APNS_PUSH_PORT));
             }
-            builder.withGatewayDestination(System.getProperty("aerogear.apns.push.host"),port);
+            builder.withGatewayDestination(System.getProperty(AEROGEAR_APNS_PUSH_HOST),port);
         }
 
         //Is the feedback gateway provided by a system property ?
-        if(System.getProperty("aerogear.apns.feedback.host")!= null){
+        if(System.getProperty(AEROGEAR_APNS_FEEDBACK_HOST)!= null){
             int port = Utilities.SANDBOX_FEEDBACK_PORT;
-            if(System.getProperty("aerogear.apns.feedback.port")!= null) {
-                port = Integer.parseInt(System.getProperty("aerogear.apns.feedback.port"));
+            if(System.getProperty(AEROGEAR_APNS_FEEDBACK_PORT)!= null) {
+                port = Integer.parseInt(System.getProperty(AEROGEAR_APNS_FEEDBACK_PORT));
             }
-            builder.withFeedbackDestination(System.getProperty("aerogear.apns.feedback.host"),port);
+            builder.withFeedbackDestination(System.getProperty(AEROGEAR_APNS_FEEDBACK_HOST),port);
         }
     }
 }
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
@@ -39,7 +39,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Set;
-import static org.jboss.aerogear.unifiedpush.message.util.SenderUtils.tryGetProperty;
+import static org.jboss.aerogear.unifiedpush.message.util.ConfigurationUtils.tryGetProperty;
 
 @SenderType(iOSVariant.class)
 public class APNsPushNotificationSender implements PushNotificationSender {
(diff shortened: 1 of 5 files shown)


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -209,4 +205,43 @@ public void messageSendFailed(ApnsNotification message, Throwable e) {
...
+    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
+        // pick the destination:
+        if (iOSVariant.isProduction()) {
+            builder.withProductionDestination();
+        } else {
+            builder.withSandboxDestination();
+        }
+
+        //Is the gateway host&port provided by a system property ?
+        if(System.getProperty("aerogear.apns.push.host")!= null){
```
matzew: can we, somehow, have test for this method ?

sebastienblanc: Yep, maybe if we could expose https://github.com/notnoop/java-apns/blob/master/src/main/java/com/notnoop/apns/ApnsService.java#L170 (that I would like to use later for the new variant wizard UI checks ) I could catch a `unknown host` or something like that

matzew: I was more talking about a unit test :-)

but this method is for sure a good one, to be integrated w/ the new UI / wizard

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -209,4 +205,43 @@ public void messageSendFailed(ApnsNotification message, Throwable e) {
...
+            int port = Utilities.SANDBOX_GATEWAY_PORT;
+            if(System.getProperty("aerogear.apns.push.port")!= null) {
```
TadeasKriz: Might be a good thing to take these strings and put them as constants at the top of this file. Firstly it would be easier to change in the future (if needed) and secondly it will be easily seen on top of the file. But I'll leave that decision to you :).

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
sebastienblanc: Yes , hard to unit test. I have added a test that can at least check that the sender is using the custom gateway we pass, tell me what you think about it.

TadeasKriz:
The code seems fine, but I'm not able to run integration tests against it to verify it.

TadeasKriz:
Btw, will this be backported into 1.0.x branch?

sebastienblanc:
@TadeasKriz yeah backport should be easy.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
sebastienblanc: Yes, make sense, refactored. thx

matzew: +1 on using a constant - thought the same when looking at the PR :-)

On Wed, Jan 21, 2015 at 4:36 PM, Tadeas Kriz <notifications@github.com>
wrote:

> In
> push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/470#discussion_r23307178>
> :
>
> > +     *  - aerogear.apns.feedback.port
> > +     * @param iOSVariant
> > +     * @param builder
> > +     */
> > +    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
> > +        // pick the destination:
> > +        if (iOSVariant.isProduction()) {
> > +            builder.withProductionDestination();
> > +        } else {
> > +            builder.withSandboxDestination();
> > +        }
> > +
> > +        //Is the gateway host&port provided by a system property ?
> > +        if(System.getProperty("aerogear.apns.push.host")!= null){
> > +            int port = Utilities.SANDBOX_GATEWAY_PORT;
> > +            if(System.getProperty("aerogear.apns.push.port")!= null) {
>
> Might be a good thing to take these strings and put them as constants at
> the top of this file. Firstly it would be easier to change in the future
> (if needed) and secondly it will be easily seen on top of the file. But
> I'll leave that decision to you :).
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/470/files#r23307178>
> .
>



-- 
Matthias Wessendorf

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -43,6 +43,10 @@
 @SenderType(iOSVariant.class)
 public class APNsPushNotificationSender implements PushNotificationSender {
 
+    private static final String AEROGEAR_APNS_PUSH_HOST = "aerogear.apns.push.host";
+    private static final String AEROGEAR_APNS_PUSH_PORT = "aerogear.apns.push.port";
+    private static final String AEROGEAR_APNS_FEEDBACK_HOST = "aerogear.apns.feedback.host";
+    private static final String AEROGEAR_APNS_FEEDBACK_PORT = "aerogear.apns.feedback.port";
```
matzew: let's name them different, like `CUSTOM_AEROGEAR....` to be CLEAR that this should be only provided under some criterias (e.g. QE tests).

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -209,4 +209,43 @@ public void messageSendFailed(ApnsNotification message, Throwable e) {
...
+    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
+        // pick the destination:
+        if (iOSVariant.isProduction()) {
+            builder.withProductionDestination();
+        } else {
+            builder.withSandboxDestination();
+        }
+
+        //Is the gateway host&port provided by a system property ?
+        if(System.getProperty(AEROGEAR_APNS_PUSH_HOST)!= null){
```
matzew: This will throw a `SecurityException` if a security manager exists and its `checkPropertyAccess()` method doesn't allow access to the specified system property.

In the past I have done these calls inside of try/catch blocks, like:
```
try {
    someProperty = System.getProperty(KEY);
    } catch (Throwable t) {
...
}
```

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -209,4 +209,43 @@ public void messageSendFailed(ApnsNotification message, Throwable e) {
...
+     */
+    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
+        // pick the destination:
+        if (iOSVariant.isProduction()) {
+            builder.withProductionDestination();
+        } else {
+            builder.withSandboxDestination();
+        }
+
+        //Is the gateway host&port provided by a system property ?
```
matzew: the above `withProductionDestination ()` or `withSandboxDestination()` is always executed, ok;

Only if these System properties are present, we override the URL/PORT. Not sure, but can we move that `System.getProperty` into a private method that is called like `checkForOverrideOfAPNsURLandPort()`? 

And its JavaDoc should be pretty clear, what's going on :-)

matzew:
One more thing - can we move this into a static block ? 

```
static {

String someurl = null;
....

try {
.....
} catch(Throwable) {...}
...
}
```

We really need to check ONCE for these System properties

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
TadeasKriz: I would suggest creating a method like so:

```
String tryGetProperty(String key) {
    try {
        return System.getProperty(key);
    } catch (SecurityException e) {
        logger.error("Could not get value of property `" + key + "` due to SecurityManager. Using null value.");
        return null; 
    }
}
```

matzew: yeah, and even move that to a Utils. in the static block, just call the
util. Makes it cleaner

On Thu, Jan 22, 2015 at 12:12 PM, Tadeas Kriz <notifications@github.com>
wrote:

> In
> push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/470#discussion_r23368736>
> :
>
> > +     * Feedback gateway and port can be override with  respectively :
> > +     *  - aerogear.apns.feedback.host
> > +     *  - aerogear.apns.feedback.port
> > +     * @param iOSVariant
> > +     * @param builder
> > +     */
> > +    private void configureDestinations(iOSVariant iOSVariant, ApnsServiceBuilder builder) {
> > +        // pick the destination:
> > +        if (iOSVariant.isProduction()) {
> > +            builder.withProductionDestination();
> > +        } else {
> > +            builder.withSandboxDestination();
> > +        }
> > +
> > +        //Is the gateway host&port provided by a system property ?
> > +        if(System.getProperty(AEROGEAR_APNS_PUSH_HOST)!= null){
>
> I would suggest creating a method like so:
>
> String tryGetProperty(String key) {
>     try {
>         return System.getProperty(key);
>     } catch (SecurityException e) {
>         logger.error("Could not get value of property `" + key + "` due to SecurityManager. Using null value.");
>         return null;
>     }
> }
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/470/files#r23368736>
> .
>



-- 
Matthias Wessendorf

TadeasKriz: :+1:

sebastienblanc:
@matzew @TadeasKriz Added a commit based on your suggestions, does this looks better ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/util/SenderUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.aerogear.unifiedpush.message.util;
+
+
+import org.jboss.aerogear.unifiedpush.utils.AeroGearLogger;
+
+public class SenderUtils {
```
matzew: I don't like the name - it's not really a util for sending

also, make the class final and give it a private ctor

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/util/SenderUtils.java
```diff
@@ -0,0 +1,40 @@
...
+     * Try to retrieve a system property and returns null if SecurityManager blocks it.
+     *
+     * @param key
+     * @return the value of the System property
+     */
+    public static String tryGetProperty(String key) {
+        try {
+            return System.getProperty(key);
+        } catch (SecurityException e) {
+            logger.severe("Could not get value of property `" + key + "` due to SecurityManager. Using null value.");
```
matzew: unfortunately the ` will be escaped - let's not use fancy chars

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -209,4 +210,43 @@ public void messageSendFailed(ApnsNotification message, Throwable e) {
...
+        if(tryGetProperty(CUSTOM_AEROGEAR_APNS_PUSH_HOST)!= null){
+            int port = Utilities.SANDBOX_GATEWAY_PORT;
+            if(tryGetProperty(CUSTOM_AEROGEAR_APNS_PUSH_PORT)!= null) {
+                port = Integer.parseInt(tryGetProperty(CUSTOM_AEROGEAR_APNS_PUSH_PORT));
+            }
+            builder.withGatewayDestination(tryGetProperty(CUSTOM_AEROGEAR_APNS_PUSH_HOST),port);
+        }
+
+        //Is the feedback gateway provided by a system property ?
+        if(tryGetProperty(CUSTOM_AEROGEAR_APNS_FEEDBACK_HOST)!= null){
```
matzew: still, this is executed all the time. As said earlier, let's move that into a static block - we really need that info once

TadeasKriz:
@sebastienblanc I think that @matzew was right about preloading the properties in a static block (well, as you implemented the tryGetProperty, you don't need a static block, but just static fields with the assignment like so:
```
private static final String customFeedbackUrl = tryGetProperty();
```

push/sender/src/test/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSenderTest.java
```diff
@@ -20,12 +20,14 @@
...
 import org.junit.Test;
-
+import static org.junit.Assert.*;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
+import java.util.Properties;
 
+import static org.junit.Assert.assertEquals;
```
matzew: two different imports of org.junit.Assert

also, let's continue using AssertJ

sebastienblanc:
@TadeasKriz @matzew Updated !

push/sender/src/test/java/org/jboss/aerogear/unifiedpush/message/util/ConfigurationUtilsTest.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.junit.Assert.*;
+import java.util.Properties;
+
+public class ConfigurationUtilsTest {
+
+    @Test
+    public void testExistingTryGetProperty(){
+        Properties properties = System.getProperties();
+        properties.setProperty("MyNiceProp","MyNiceValue");
+        assertEquals(ConfigurationUtils.tryGetProperty("MyNiceProp"),"MyNiceValue");
```
matzew: LEt's continue using AssertJ

sebastienblanc:
@matzew converted to assertJ

matzew:
thanks, landed in master

and also backported to 1.0.x

PR #482, opened 2015-01-29
Adding parent, containing latest KC 1.1.0.Final

matzew:
You need https://github.com/aerogear/aerogear-parent/pull/40 for testing this

Tested and deployed on WF 8.2 -> works fine

NOTE: before merging this, I'd think we should get the parent 0.2.12 released to maven central


matzew:
@abstractj Let me start with the 0.2.12 release

PR #483, opened 2015-01-30
Removal of ChromePackaged App classes and UI.  Add a deprecation notice ...

lholmquist:
This is for https://issues.jboss.org/browse/AGPUSH-928

Since the chrome.pushMessaging api is going away very soon in favor of just using the same gcm for android, so  the way we do chrome apps on the UPS will be wrong, so this removes all revelant chrome packaged app stuff.

There is a new updated Guide in this branch, https://github.com/lholmquist/aerogear.org/tree/deprecate-chrome
which i will PR once this has been added


README.md
```diff
@@ -100,6 +100,20 @@ The content of the [Release Process](https://github.com/aerogear/collateral/wiki
 mvn release:GOAL -Pdist,test
 ```
 
+
+## Deprecation Notices
```
matzew: let's make sure we have this part of the release notes as well

matzew:
Can you rebase this, Luke ?

lholmquist:
@matzew rebased

PR #486, opened 2015-02-02
APNs sender should support both iOS and Safari - AGPUSH-491

lholmquist:
Putting the time in to create an APNsVariant to combine iOS and Safari is not really worth it anymore.  We can just re-user the iOS Variant similar to how we use the Android Variant for chrome apps.

However; the message format is slightly different for Safari,  so the APNs sender needs to be updated

### DIFF
--- a/push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/Message.java
+++ b/push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/Message.java
@@ -32,7 +32,12 @@ public class Message {
     @JsonProperty("action-category")
     private String actionCategory;
     private String alert;
+    private String title;
+    private String action;
     private String sound;
+
+    @JsonProperty("url-args")
+    private String[] urlArgs;
 
     @JsonProperty("content-available")
     private boolean contentAvailable;
@@ -71,6 +76,26 @@ public class Message {
     public void setAlert(String alert) {
         this.alert = alert;
     }
+
+    /**
+     * Returns the value of the 'title' key from the submitted payload.
+     * This key is recognized in APNs for Safari, without any API invocation and
+     * on AeroGear's GCM offerings.
+     *
+     */
+    public String getTitle() { return title; }
+
+    public void setTitle(String title) { this.title = title; }
+
+    /**
+     * Returns the value of the 'action' key from the submitted payload.
+     * This key is recognized in APNs for Safari, without any API invocation and
+     * on AeroGear's GCM offerings.
+     *
+     */
+    public String getAction() { return action; }
+
+    public void setAction(String action) { this.action = action; }
 
     /**
      * Returns the value of the 'sound' key from the submitted payload.
@@ -132,6 +157,16 @@ public class Message {
     public void setUserData(Map<String, Object> userData) {
         this.userData = userData;
     }
+
+    /**
+     * Returns the value of the 'url-args' key from the submitted payload.
+     * This key is recognized in APNs for Safari, without any API invocation and
+     * on AeroGear's GCM offerings.
+     *
+     */
+    public String[] getUrlArgs() { return urlArgs; }
+
+    public void setUrlArgs(String[] urlArgs) { this.urlArgs = urlArgs; }
 
     /**
      * Returns the SimplePush specific version number.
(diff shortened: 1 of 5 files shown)


matzew:
The message format should be also documented here, since that is used for API doc generation
https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java#L52-L80

matzew:
What happens when one uses a development profile w/ Safari ?

lholmquist:
@matzew wrt to using a dev profile with Safari, i would imagine the same thing if you used the dev profile with a production certificate on iOS(assumes this doesn't work on iOS)

lholmquist:
@matzew updated the API doc

matzew:
> i would imagine the same thing if you used the dev profile with a production certificate on iOS(assumes this doesn't work on iOS)

can you test? I don't want the server to explode, just because lack of manual testing.

The doc will say that you have to have production cert, that's OK. I hope using a development cert causes no harm, to the server.

lholmquist:
in the apple dev center when setting up the certificate,  there is only the option for a Production certificate.

I will test however, using a production certificate, and selecting the development radio button

matzew:
yeah, that' what I meant - sorry :-)

push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/UnifiedPushMessage.java
```diff
@@ -33,6 +33,9 @@
  * <pre>
  *  "message": {
  *   "alert": "HELLO!",
+ *   "title": "Safari Title",
```
matzew: It's actually not just Safari title - Apple Watch will use that too (and I think iPhone etc as well, with iOS 8.2)

matzew:
@lholmquist do we have a JIRA to update the Sender libs for Java and Node to refelect this ?

lholmquist:
@matzew let me check and create if not

push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/UnifiedPushMessage.java
lholmquist: right,  let me update

lholmquist:
@matzew There is https://issues.jboss.org/browse/AGPUSH-1260 and  https://issues.jboss.org/browse/AGPUSH-1261 however,  i think it is just for the "title-x" keys.  Probably ok to update the JIRA so it also gets the safari stuff

lholmquist:
@matzew updated and rebased

push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/Message.java
```diff
@@ -32,8 +32,13 @@
     @JsonProperty("action-category")
     private String actionCategory;
     private String alert;
+    private String title;
+    private String action;
     private String sound;
 
+    @JsonProperty("url-args")
+    private String[] urlArgs;
```
matzew: Oh, just noticed this, what's the reason not using varargs here ?

matzew: I mean on the the setter

lholmquist: @matzew JS guy writing Java

matzew: ok :-)

merging

matzew:
landed https://github.com/aerogear/aerogear-unifiedpush-server/commit/7750a494bc8b908f568ce28be23119ca17d2a772

matzew:
@lholmquist the APNs variant is still named iOS variant, and mentions:

```
Apple's Push Network (APNs) will be used. To learn more about APNs, visit our iOS or Apache Cordova guides for push.
```

This does not sound like I could do Safari here. Do we want to keep that ? and just have a doc  ? Or would be more pro-active and actually give the user a hint that this will work for Safari as well ?

lholmquist:
@matzew that is part of this: https://issues.jboss.org/browse/AGPUSH-456

once we have a doc, then we can link to it,  similar to the android/
chrome one

PR #500, opened 2015-02-19
Make use of a separate datasource for Keycloak

abstractj:
Like [reported](http://lists.jboss.org/pipermail/keycloak-dev/2015-February/003682.html) by @sebastienblanc at keycloak-dev, there's an issue after the deployment of Keycloak 1.1.0.Final and UPS pointing to the same database.

Marek [suggested](http://lists.jboss.org/pipermail/keycloak-dev/2015-February/003685.html) to split the datasources, which makes perfect sense.

That said @matzew @edewit @qmx @sebastienblanc would you mind to review?

It must go on `master` and `1.0.x` branch.

### DIFF
--- a/databases/h2-database-config-wildfly.cli
+++ b/databases/h2-database-config-wildfly.cli
@@ -5,5 +5,8 @@ batch
 ## Add UnifiedPush Datasource
 data-source add --name=UnifiedPushDS --driver-name=h2 --jndi-name=java:jboss/datasources/UnifiedPushDS --connection-url="jdbc:h2:${jboss.server.data.dir}/unifiedpush;DB_CLOSE_DELAY=-1" --user-name=sa --password=sa --use-ccm=true --enabled=true
 
+## Add Keycloak Datasource
+data-source add --name=KeycloakDS --driver-name=h2 --jndi-name=java:jboss/datasources/KeycloakDS --connection-url="jdbc:h2:${jboss.server.data.dir}/keycloak;DB_CLOSE_DELAY=-1" --user-name=sa --password=sa --use-ccm=true --enabled=true
+
 run-batch
 #:reload
--- a/databases/h2-database-config.cli
+++ b/databases/h2-database-config.cli
@@ -6,5 +6,11 @@ batch
 data-source add --name=UnifiedPushDS --driver-name=h2 --jndi-name=java:jboss/datasources/UnifiedPushDS --connection-url="jdbc:h2:${jboss.server.data.dir}/unifiedpush;DB_CLOSE_DELAY=-1" --user-name=sa --password=sa --use-ccm=true
 data-source enable --name=UnifiedPushDS
 
+## Add Keycloak Datasource
+data-source add --name=KeycloakDS --driver-name=h2 --jndi-name=java:jboss/datasources/KeycloakDS --connection-url="jdbc:h2:${jboss.server.data.dir}/keycloak;DB_CLOSE_DELAY=-1" --user-name=sa --password=sa --use-ccm=true
+data-source enable --name=KeycloakDS
+
+
+
 run-batch
 #:reload
--- a/databases/mysql-database-config-wildfly.cli
+++ b/databases/mysql-database-config-wildfly.cli
@@ -8,5 +8,8 @@ batch
 ## Add UnifiedPush Datasource
 data-source add --name=UnifiedPushDS --driver-name=mysqlup --jndi-name=java:jboss/datasources/UnifiedPushDS --connection-url="jdbc:mysql://localhost:3306/unifiedpush?useUnicode=true&amp;characterEncoding=UTF-8" --user-name=unifiedpush --password=unifiedpush --use-ccm=false --max-pool-size=25 --blocking-timeout-wait-millis=5000 --enabled=true
 
+## Add Keycloak Datasource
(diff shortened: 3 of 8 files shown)


qmx:
Schweeet!

the only caveat is that I think it would be wise to do this change on master only, since this has direct impact on the migration work. (this would mean that for 1.0.x users would need to provide two datasources) + we would need to migrate data from our DS to the new DS (doable, but hard work)

qmx:
> (this would mean that for 1.0.x users would need to provide two datasources)

Gaah, confusing comment is confusing:

for the migrator, user needs to provide DB credentials - which essentially means we would need to create a different migrator, one for UPS and another one for keycloak.

sebastienblanc:
I will make sure tomorrow checking if this issue is reproducible on 1.0.x , that will make the decision on what to do easier.

Envoyé de mon iPhone

sebastienblanc:
For MySQL and PostGres, the database also needs to be created manually  like mentioned here https://aerogear.org/docs/unifiedpush/ups_userguide/server-installation/#_mysql_database (section 3) .  tracked by https://issues.jboss.org/browse/AGPUSH-1292

sebastienblanc:
Ok, tested on 1.0.x, we don't have the issue so since we are on the critical path for the 1.0.3 I would suggest to do nothing for this branch,
But, let me be completly honest, I just failed to reproduce that issue @lholmquist  and me were having, not sure sure if something happened since then (new parent or smt else)  ... However, I think this is still a nice enhancement having these 2 in a separate databse but only for master.

matzew:
+1 let's not touch 1.0.x with this

On Fri, Feb 20, 2015 at 10:20 AM, Sebastien Blanc <notifications@github.com>
wrote:

> Ok, tested on 1.0.x, we don't have the issue so since we are on the
> critical path for the 1.0.3 I would suggest to do nothing for this branch,
> But, let me be completly honest, I just failed to reproduce that issue
> @lholmquist <https://github.com/lholmquist> and me were having, not sure
> sure if something happened since then (new parent or smt else) ... However,
> I think this is still a nice enhancement having these 2 in a separate
> databse but only for master.
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/500#issuecomment-75209791>
> .
>



-- 
Matthias Wessendorf

abstractj:
@qmx @sebastienblanc @matzew ok, so cherry-pick to master?

abstractj:
@sebastienblanc btw, not merging it, from my pov is a blocker to decouple Keycloak from UPS.

sebastienblanc:
@abstractj sure, I was also thinking that this would be a good one for the decouple task. I'm totally +1 to have this on master.

qmx:
I'm eager to hear ideas on how to provide a migration path from 1.0.x

matzew:
> I'm eager to hear ideas on how to provide a migration path from 1.0.x

@qmx I understood that the migration work does include going from 1.0.x to 1.1.0 as well

qmx:
@matzew there's no miracle as in I can ensure schema changes inside the same DS - for us to move keycloak's data over another datasource it needs to be something else

matzew:
ok, so... I think, having the separation on 1.1.0/master only may be the best.

That way we can have 'smooth' migrations w/in 1.0.x series.

But not sure what's exactly the best migration path from 1.0.3 -> 1.1.0 (with separated datasources)

abstractj:
So, what is the consensus here?

sebastienblanc:
From what I understand , only for master and I really want to test that and provide feedback before the end of the week.

Envoyé de mon iPhone

matzew:
I *think* that we should get it on to master. I am not sure for 1.0.x and
our last 1.0.3 release.

At some (later) point we do need to think about going from 1.0.3 to 1.1.x.

But I am not sure we need to separated the DS for our last 1.0.3.

@qmx @sebastienblanc what are you guys thinking?

On Wed, Feb 25, 2015 at 7:36 PM, Bruno Oliveira <notifications@github.com>
wrote:

> So, what is the consensus here?
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/500#issuecomment-76026189>
> .
>



-- 
Matthias Wessendorf

sebastienblanc:
@matzew Yes let's keep it for master only

sebastienblanc:
Ok, I tested the PR : 
- With H2 :+1: 
- With MySql :+1:  but I had to run some extra sql commands : 
```
 create database keycloak default character set = "UTF8" default collate = "utf8_general_ci";
GRANT SELECT,INSERT,UPDATE,ALTER,DELETE,CREATE,DROP ON keycloak.* TO 'keycloak'@'localhost';

```

sebastienblanc:
For PostGres, I'm currently having really hard time to reinstall it on my fedora, so if anyone want to test it please :)

abstractj:
@matzew @qmx @sebastienblanc landed on master, right here https://github.com/aerogear/aerogear-unifiedpush-server/commit/55e4725d3afcf51f16c7b8a4c1ccde860a9f54b8

sebastienblanc:
Cool but has anyone tested that with postgresql ?

abstractj:
@sebastienblanc yup, but if that only works on my machine. `git revert` to the rescue ;)

sebastienblanc:
@abstractj Ah cool ! I totally trust you :) , let me PR now for  https://issues.jboss.org/browse/AGPUSH-1292

PR #505, opened 2015-02-24
AGPUSH-1295 : fix remove installations with shared categories

sebastienblanc:
@TadeasKriz @edewit Could you review ?

### DIFF
--- a/model/jpa/src/main/resources/META-INF/orm.xml
+++ b/model/jpa/src/main/resources/META-INF/orm.xml
@@ -48,7 +48,7 @@ http://java.sun.com/xml/ns/persistence/orm_2_0.xsd"
             </many-to-one>
             <many-to-many name="categories" fetch="EAGER">
                 <cascade>
-                    <cascade-all/>
+                    <cascade-persist/>
                 </cascade>
             </many-to-many>
         </attributes>


sebastienblanc:
This needs also to be cherry-picked to master.

service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/ClientInstallationServiceImpl.java
```diff
@@ -168,6 +168,10 @@ public Installation findById(String primaryKey) {
 
     @Override
     public void removeInstallation(Installation installation) {
+        //Let manually remove the categories and update the installations before deleting
+        //check AGPUSH-1295 for more information
+        installation.setCategories(null);
+        updateInstallation(installation);
```
edewit: Instead of manually updating the categories before deleting couldn't we change the cascade setting?
https://github.com/aerogear/aerogear-unifiedpush-server/blob/159c119c6ef1564846b847e6b418335c2ce01dc0/model/jpa/src/main/resources/META-INF/orm.xml#L51
to
```xml
<cascade>
   <cascade-persist/>
</cascade>
```

sebastienblanc: Let me try that

sebastienblanc:
@edewit you are absolutely right , updated the PR

edewit:
:+1:

sebastienblanc:
landed on 1.0.x e34f6a62e633d111a6dde3586be40bd3dd835261 on master 11a6e8b8b399acbb58a42999ef10966b786145e6

matzew:
@sebastienblanc do we have a test for that ? e.g. one that shows the bug? and now runs through (green bar) ?

sebastienblanc:
By replacing (and forcing) my previous commit with edewit suggestion, I
forgot to push again the test I added to the service test class. I can add
it but as mentioned on IRC, the test case goes green even before the fix,
not sure what happens here, maybe OpenEJB ? Maybe an Arquillian test could
fix that.

matzew:
How is that related to OpenEJB ? the model (jpa) tests have no EJB container at all.

sebastienblanc:
I don't know just something that popped out of my brain :) , here is the
test I had https://gist.github.com/sebastienblanc/afe5629873c2182848e8

matzew:
but that's a service test - I mean, a model (jpa) test for that

On Tue, Feb 24, 2015 at 6:50 PM, Sebastien Blanc <notifications@github.com>
wrote:

> I don't know just something that popped out of my brain :) , here is the
> test I had https://gist.github.com/sebastienblanc/afe5629873c2182848e8
>
> On Tue, Feb 24, 2015 at 6:49 PM, Matthias Wessendorf <
> notifications@github.com> wrote:
>
> > How is that related to OpenEJB ? the model (jpa) tests have no EJB
> > container at all.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/aerogear/aerogear-unifiedpush-server/pull/505#issuecomment-75807919
> >
>
> > .
> >
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/505#issuecomment-75808224>
> .
>



-- 
Matthias Wessendorf

sebastienblanc:
I had one as well with same effect (green without the fix) let me PR them,
can not hurt ...

sebastienblanc:
@matzew Test updated for the JPA and test added for the service in this PR https://github.com/aerogear/aerogear-unifiedpush-server/pull/506

PR #507, opened 2015-02-27
adding logback + slf4j logging

IvanGurtler:
testing on EAP6.3 (wildfly  problems can be solved by jboss-deployment-structure.xml  ... but not tested on wildfly)

PR #508, opened 2015-03-05
Logger

matzew:
updated version of https://github.com/aerogear/aerogear-unifiedpush-server/pull/507


matzew:
@sebastienblanc will open a different one

PR #512, opened 2015-03-12
Hibernate.hbm2ddl back to original value

matzew:
Fixes https://issues.jboss.org/browse/AGPUSH-1309 
@qmx @sebastienblanc please review/merge

### DIFF
--- a/model/jpa/src/main/resources/META-INF/persistence.xml
+++ b/model/jpa/src/main/resources/META-INF/persistence.xml
@@ -25,7 +25,7 @@
     <mapping-file>META-INF/orm.xml</mapping-file>
 
     <properties>
-      <property name="hibernate.hbm2ddl.auto" value="validate"/>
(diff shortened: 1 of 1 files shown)


sebastienblanc:
Tested with EAP 6.3 : 
* Clean deploy with H2 :+1: 
* Migration 1.0.0 > 1.0.3 MySQL :+1: 
* Migration 10.2 > 1.0.3 Mysql :+1: 

so, I'm :+1:

qmx:
if you turn this option back, you are going to lose db schema versioning and by consequence have two things fiddling with DDL statements. Later in time you'll wonder why some data is missing after a migration, and the cause is that hibernate went and created a new empty table after you renamed one of your entities, without caring for the existing data & relationships.

I'm closing this.

matzew:
agreed - closed the JIRA as well

PR #521, opened 2015-04-03
add new user creation script and enable full scope for oauth client

sebastienblanc:
* Script that shows how to create an user
* Allow full scope for the oauth client (to have user creation rights)

BTW, I tested the scripts against https (on openshift) and it works, https node mod must be used instead of http.

@matzew can you review ?

### DIFF
--- a/servers/auth-server/src/main/webapp/WEB-INF/ups-realm.json
+++ b/servers/auth-server/src/main/webapp/WEB-INF/ups-realm.json
@@ -25,7 +25,7 @@
             ],
             "realmRoles": [ "admin" ],
             "applicationRoles": {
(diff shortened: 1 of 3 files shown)


servers/auth-server/src/main/webapp/WEB-INF/ups-realm.json
```diff
@@ -25,7 +25,7 @@
             ],
             "realmRoles": [ "admin" ],
             "applicationRoles": {
-               "realm-management": [ "realm-admin" ],
+               "realm-management": [ "realm-admin"],
```
matzew: formatting issue

PR #529, opened 2015-04-15
remove detached entity

matzew:
loading the entity if not part of the transaction before trying to delete

### DIFF
--- a/model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java
+++ b/model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java
@@ -62,6 +62,12 @@ public abstract class JPABaseDao<T, K> implements GenericBaseDao<T, K> {
 
     public void delete(T entity) {
         if (entity != null) {
+            // making sure the entity in question,
+            // is really part of this transaction
+            if (! entityManager.contains(entity)) {
+                entity = entityManager.merge(entity);
+            }
+
             entityManager.remove(entity);
         }
     }


model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java
```diff
@@ -62,6 +62,12 @@ public void update(T entity) {
 
     public void delete(T entity) {
         if (entity != null) {
+            // making sure the entity in question,
+            // is really part of this transaction
+            if (! entityManager.contains(entity)) {
+                entity = entityManager.merge(entity);
```
edewit: This will load the entity from the database, better would be:
```java
entityManager.getReference(getType(), entity.getId());
```

sebastienblanc: Please be sure we test this also with installations associated to
categories (and categories that are shared by other installations as well)

matzew: @sebastienblanc they stay :-) I checked that ;-)

matzew: @edewit I have no info on the .getId() - instead of casting I did this.

However, I can see this a nice improvement. OK if we file a JIRA for 1.1.x?

edewit: sure @matzew guess we only need to say `T extends BaseEntity` or something like that, but let's do that later

matzew: that's also conficting w/ changes done from Lukas on the JMS.

I will file a JIRA for future evaluation of improving this

On Wed, Apr 15, 2015 at 2:57 PM, Erik Jan de Wit <notifications@github.com>
wrote:

> In
> model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/529#discussion_r28416322>
> :
>
> > @@ -62,6 +62,12 @@ public void update(T entity) {
> >
> >      public void delete(T entity) {
> >          if (entity != null) {
> > +            // making sure the entity in question,
> > +            // is really part of this transaction
> > +            if (! entityManager.contains(entity)) {
> > +                entity = entityManager.merge(entity);
>
> sure @matzew <https://github.com/matzew> guess we only need to say T
> extends BaseEntity or something like that, but let's do that later
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/529/files#r28416322>
> .
>



-- 
Matthias Wessendorf

PR #542, opened 2015-05-05
Split UnifiedPushMessage to hold internal information

idelpivnitskiy:
JIRA issue: https://issues.jboss.org/browse/AGPUSH-1381

@matzew @sebastienblanc I've split `UnifiedPushMessage` and (for the quick fix) I've removed `ipAddress` from test json files. Is it ok? Should I override `UnifiedPushMessage.toStrippedJsonString()`?

### DIFF
--- a/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
+++ b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
@@ -28,8 +28,8 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
 import org.jboss.aerogear.unifiedpush.api.PushApplication;
+import org.jboss.aerogear.unifiedpush.message.InternalUnifiedPushMessage;
 import org.jboss.aerogear.unifiedpush.message.NotificationRouter;
-import org.jboss.aerogear.unifiedpush.message.UnifiedPushMessage;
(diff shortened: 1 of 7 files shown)


push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/InternalUnifiedPushMessage.java
```diff
@@ -0,0 +1,26 @@
+package org.jboss.aerogear.unifiedpush.message;
```
matzew: I wonder if we should move this class out of the push/model package.

That API is used on the Java Sender, and there it does not make sense to have these extra information (like IP address, and others).

Perhaps it's good to move it to push/sender module?

matzew: Let's also move the `clientIdentifier` into this class:
https://github.com/idelpivnitskiy/aerogear-unifiedpush-server/blob/AGPUSH-1381/push/model/src/main/java/org/jboss/aerogear/unifiedpush/message/UnifiedPushMessage.java#L83

matzew:
@idelpivnitskiy @sebastienblanc Once this is in, I will rebase https://github.com/aerogear/aerogear-unifiedpush-server/pull/543 and leverage this new Internal Message class for some of the push message ID etc

matzew:
:+1: awesome, thanks @idelpivnitskiy

matzew:
Landed in 3d749a8e5ba3bd00d0b6decd019664a095cc06e5

idelpivnitskiy:
@matzew added copyright to my IDEA settings :) thanks!

PR #540, opened 2015-04-29
AGPUSH-971 : Advanced Analytics, Server Side 

sebastienblanc:
New PR Based on the closed https://github.com/aerogear/aerogear-unifiedpush-server/pull/523

* rebase afer JMS landed on master branch :white_check_mark: 
* `request.getHeader("push-identifier");` header needs a different prefix (aerogear-push-id) :white_check_mark: 
* move the `extractPushIdentifier()` to the `HttpRequestUtil` util :white_check_mark: 
* first/last opened will stay as is, right ? :white_check_mark: 
* `message.getMessage().getUserData().put("push-identifier",pushMessageInformation.getId());` needs to be done at outer/different/custom level - not on the envelope/payload (e.g. the user data)  :white_check_mark:  (has been removed from "our" user-data but for each push network I can only end up in their respective "custom" section)

* leverage the MetricsCollector instead of new VariantMetricInformationDao on this PR : *Will be handled  when Registration will use messaging*
* push identifier: docment it on the endpoint doc : *REST Endpoint has been updated but update to the endpoint doc will be in a aerogear.org PR*
* JavaDoc on `NotificationSenderCallback` : *not relevant anymore*

* `pushMessageInformation.setAppOpenCounter` needs atomicLong or AtomicRef : *still open for discussion, we have now the "simple" version (incrementing a long)*

@lfryc @edewit @lfryc can you take a look ?

PR #543, opened 2015-05-05
Analytics

matzew:
updated version of https://github.com/aerogear/aerogear-unifiedpush-server/pull/540


model/api/src/main/java/org/jboss/aerogear/unifiedpush/dao/PushMessageInformationDao.java
```diff
@@ -114,4 +114,7 @@
     List<String> findVariantIDsWithWarnings();
     List<PushMessageInformation> findLatestActivity(int maxResults);
     long getNumberOfPushMessagesForApplications();
+
+    PushMessageInformation getPushMessageInformation(String id);
```
matzew: @sebastienblanc what is the ID here, the PK ?

model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPAPushMessageInformationDao.java
```diff
@@ -141,6 +141,13 @@ public long getNumberOfPushMessagesForApplications() {
                 "IN (select p.pushApplicationID from PushApplication p)", Long.class).getSingleResult();
     }
 
+    @Override
+    public PushMessageInformation getPushMessageInformation(String id) {
+        return getSingleResultForQuery(createQuery(
+                "select pmi from PushMessageInformation pmi where pmi.id = :id")
+                .setParameter("id", id));
+
```
matzew: @sebastienblanc if that is the PK, why not using `em.find(PushMessageInformation.class, id);` ?

edewit: good find, but even better would be to replace it with [`find(id)`](https://github.com/aerogear/aerogear-unifiedpush-server/blob/d216bfa1a9a4fe9a9165874cec0ed79b79ab80d6/model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java#L60) so this entire method could be removed

matzew: ok, wil do

On Wed, May 6, 2015 at 3:06 PM, Erik Jan de Wit <notifications@github.com>
wrote:

> In
> model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPAPushMessageInformationDao.java
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/543#discussion_r29755713>
> :
>
> > @@ -141,6 +141,13 @@ public long getNumberOfPushMessagesForApplications() {
> >                  "IN (select p.pushApplicationID from PushApplication p)", Long.class).getSingleResult();
> >      }
> >
> > +    @Override
> > +    public PushMessageInformation getPushMessageInformation(String id) {
> > +        return getSingleResultForQuery(createQuery(
> > +                "select pmi from PushMessageInformation pmi where pmi.id = :id")
> > +                .setParameter("id", id));
> > +
>
> good find, but even better would be to replace it with find(id)
> <https://github.com/aerogear/aerogear-unifiedpush-server/blob/d216bfa1a9a4fe9a9165874cec0ed79b79ab80d6/model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPABaseDao.java#L60>
> so this entire method could be removed
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/543/files#r29755713>
> .
>



-- 
Matthias Wessendorf

matzew:
@corinnekrych @edewit @danielpassos w/ https://github.com/matzew/aerogear-unifiedpush-server/commit/86134ea31f563f662441aabd4cef3107acba6594 we have a PUT endpoint for storing the "app opened due to push" ack.

A little CURL I did is:

```
curl -u "super:secret" -v -H "Accept: application/json" -H "Content-type: application/json" -X PUT http://127.0.0.1:8080/ag-push/rest/registry/device/pushMessage/eac8c56c-209e-4f20-95e4-c59404947bf6 
```

I have the endpoint under /registry/device for a few reasons:
* it's data sent from the actual device
* and no need to add another http basic exception

matzew:
@secondsun ^ 

Now that we have a separate endpoint for the metrics - do you think we need a `PUT` for "updating" the registration metadata? or should the mobile clients be executing against our `POST` registration API for both: initial registration and continued/updated registration ?

secondsun:
@matzew That kind of depends ;)

Doing a bit of research (IE bothering you in IRC) on Android and SimplePush the device installation id used by the server is provided by the client.  On Android we repeatedly call .register (because of a laundry list of technical reasons).  This invokes a POST request every time.  Because we know the ID and because multiple calls to register always leave the server in the same state we should probably be using PUT for both registration AND updating.

This is mostly a round about way of saying I think we should not create a second method for android setting updates and possibly, in the future, switch to using PUT instead of POST.

matzew:
@secondsun given the amount of tickets we have for 1.1.0, I'd vote to not do that change for 1.1.0

I see your point, but being pragmatic it does not hurt to do POST for both: initial registration and updating registration data. 

I have filed a future JIRA: https://issues.jboss.org/browse/AGPUSH-1399

secondsun:
@matzew  +1 for pragmatism

cvasilak:
I am +1 for the separate endpoint for the metric analytic update. 
regarding the POST/PUT reg update can be looked later on I guess

:+1:

danielpassos:
:+1:  for the separate endpoint

PR #548, opened 2015-05-08
Updating CURL / Java Senders

matzew:
@lfryc @lholmquist please review. If needed, send a Node.js PR against this branch ;-)


lholmquist:
@matzew sent one, https://github.com/aerogear/aerogear-unifiedpush-server/pull/551


and +1 for this

lfryc:
:eyes:

lfryc:
Java sender works perfectly, but when I copy&pasted curl command, it can't be either executed in the shell as it is or copied into shell script and get executed.

I believe we should add back slashes `\` at the end of each line for curl command, shouldn't we?

matzew:
@lfryc rebased and besides shrinking to a usful minimum, I have added the backslashes

lfryc:
:+1: @matzew yep, all senders works perfectly now!

merging...

PR #551, opened 2015-05-08
Update node.js sender snippet

lholmquist:


lfryc:
:eyes:

lfryc:
works fine, but it is merged anyway :-)

PR #554, opened 2015-05-12
improve Send Notification dialog AGPUSH-1389

lfryc:
https://issues.jboss.org/browse/AGPUSH-1389

Testing:

* click 'Send Notification to this app'
 * verify that close button works
 * [interactive combo-box should be visible](http://andresgalante.com/ups-console/app-detail-variants.html) instead of standard `<select multiple />` combo box


lfryc:
@andresgalante I did some small changes to LESS in order to avoid need for another icon set (glypicons) when the one we used already (font-awesome) can be used.

lfryc:
unrelated test failure

matzew:
:+1:

sebastienblanc:
:+1:

PR #557, opened 2015-05-15
Service reuse and batch size

lfryc:
[AGPUSH-1365](https://issues.jboss.org/browse/AGPUSH-1365)	Reuse existing APNs socket for all batches triggered by one push message
[AGPUSH-1357](https://issues.jboss.org/browse/AGPUSH-1357)	Allow to configure TokenLoader's batch size for each network type separately
[AGPUSH-1349](https://issues.jboss.org/browse/AGPUSH-1349) Write unit tests for batch loading and JMS processing

Motivation:
* `alpha2` state of UPS messaging was suboptimal in terms of cooperation with services such as ApnsService
  * ApnsService requires pushing a large number of tokens in one batch
  * and reuse of sockets between each particular push
* in order to implement reuse of sockets, we had to know when the pushing was actually completed (so that the service could be teared down)
  * I added few more numbers information to `PushMessageInformation/VariantMetricInformation` to track served/total numbers
  * added auxiliary messaging that helps `MetricCollector` decide when push for one pushmessage and one variant is completed  

Testing:
* please test delivery to iOS and Android devices
* functionality is partially covered by unit tests, but they does not cover integration of all components used together
* if you register two or more iOS devices, you should be able to emulate delivery in two batches via configuring `./bin/standalone.sh -c standalone-full.xml -b 0.0.0.0 -Daerogear.ios.batchSize=1` - the channel for APNs sending should be reused between those 2


model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/PushMessageInformation.java
```diff
@@ -175,6 +180,22 @@ public void setLastOpenDate(Date lastOpenDate) {
         this.lastOpenDate = lastOpenDate;
     }
 
+    public int getServedVariants() {
```
lfryc: javadoc

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/VariantMetricInformation.java
```diff
@@ -115,4 +119,20 @@ public void setVariantOpenCounter(long variantOpenCounter) {
     public void incrementVariantOpenCounter() {
         this.setVariantOpenCounter(this.getVariantOpenCounter() + 1);
     }
+
+    public int getServedBatches() {
```
lfryc: javadoc

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/PushMessageInformation.java
```diff
@@ -175,6 +180,22 @@ public void setLastOpenDate(Date lastOpenDate) {
...
+        return servedVariants;
+    }
+
+    public void setServedVariants(int servedVariants) {
+        this.servedVariants = servedVariants;
+    }
+
+    public int getTotalVariants() {
```
matzew: javadoc ;-)

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/NotificationDispatcher.java
```diff
@@ -69,7 +69,7 @@ public void sendMessagesToPushNetwork(@Observes @Dequeue MessageHolderWithTokens
 
         logger.fine("Received message from queue: " + msg.getUnifiedPushMessage().getMessage().getAlert());
 
-        senders.select(new SenderTypeLiteral(variant.getClass())).get()
+        senders.select(new SenderTypeLiteral(variant.getType())).get()
```
matzew: :eyes:

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/AbstractServiceCache.java
```diff
@@ -0,0 +1,237 @@
...
+
+    private void freeUpHolder(InstanceKey key, Holder holder) {
+        holderMap.remove(key, holder);
+    }
+
+    public static interface ServiceConstructor<T> {
+        T construct();
+    }
+
+    private class Holder {
```
matzew: can you add a bit of doc here? holding what, why?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/AbstractServiceCache.java
```diff
@@ -0,0 +1,237 @@
...
+                        }
+                    }
+                    return service;
+                }
+            }
+            return null;
+        }
+    }
+
+    private static class InstanceKey {
```
matzew: same

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/ApnsServiceCache.java
```diff
@@ -0,0 +1,66 @@
...
+ * @see AbstractServiceCache#dequeueOrCreateNewService(String, org.jboss.aerogear.unifiedpush.api.Variant, org.jboss.aerogear.unifiedpush.message.cache.AbstractServiceCache.ServiceConstructor)
+ * @see AbstractServiceCache#queueFreedUpService(String, org.jboss.aerogear.unifiedpush.api.Variant, Object)
+ * @see AbstractServiceCache#freeUpSlot(String, org.jboss.aerogear.unifiedpush.api.Variant)
+ */
+@ApplicationScoped
+public class ApnsServiceCache extends AbstractServiceCache<ApnsService> {
+
+    private final AeroGearLogger logger = AeroGearLogger.getInstance(ApnsServiceCache.class);
+
+    public static final int INSTANCE_LIMIT = 15;
```
matzew: let's do 10

push/sender/src/test/java/org/jboss/aerogear/unifiedpush/test/archive/UnifiedPushArchiveBase.java
```diff
@@ -0,0 +1,190 @@
...
+     * {@inheritDoc}
+     *
+     * @see org.jboss.shrinkwrap.impl.base.container.WebContainerBase#getWebInfPath()
+     */
+    @Override
+    protected ArchivePath getServiceProvidersPath() {
+        return PATH_SERVICE_PROVIDERS;
+    }
+
+    public UnifiedPushArchive withMockito() {
```
lfryc: let's remove `return null` methods from abstract class ;-) leave it for final impl

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/exception/PushNetworkUnreachableException.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Thrown when {@link PushNotificationSender} failed to establish connection to the Push Network.
+ *
+ * In such a case, the message will be considered not dispatched and messaging subsystem will try to deliver message later.
+ *
+ * Note: this exception should NOT be thrown when the variant setup is incorrect and Push Network refused connection (e.g.
+ * because of failed authentication)
+ */
+public class PushNetworkUnreachableException extends DispatchInitiationException {
```
matzew: right now this is only thrown for apns?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+    private SenderConfiguration loadConfigurationFor(VariantType type, SenderConfiguration defaultConfiguration) {
+        return new SenderConfiguration(
+                getProperty(type, "batchesToLoad", defaultConfiguration.batchesToLoad(), Integer.class),
+                getProperty(type, "batchSize", defaultConfiguration.batchSize(), Integer.class)
+            );
+    }
+
+    private <T> T getProperty(VariantType type, String property, T defaultValue, Class<T> expectedType) {
+        String systemPropertyName = String.format("aerogear.%s.%s", type.getTypeName(), property);
+        String value = System.getProperty(systemPropertyName, defaultValue.toString());
```
matzew: can we use https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/util/ConfigurationUtils.java ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/exception/PushNetworkUnreachableException.java
matzew: I think it would be nice, if this class would be used in https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/util/HealthNetworkServiceImpl.java

matzew: perhaps that's a worth a refactoring JIAR ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/AllBatchesLoaded.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.aerogear.unifiedpush.message.holder;
+
+import java.io.Serializable;
+
+/**
+ * Event fired when all batches for given variant are loaded and dispatched for further processing
+ */
```
matzew: so, after being done, for the variant ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/BatchLoaded.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package org.jboss.aerogear.unifiedpush.message.holder;
+
+import java.io.Serializable;
+
+/**
+ * Event fired when one batch for given variant was loaded and dispatched for further processing
+ */
+public class BatchLoaded implements Serializable {
```
matzew: for one, of n batches ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/VariantCompleted.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.jboss.aerogear.unifiedpush.message.holder;
+
+import org.jboss.aerogear.unifiedpush.message.MetricsCollector;
+
+/**
+ * Event fired by {@link MetricsCollector} when given variant was completely served when dispatching message for given push message
+ */
+public class VariantCompleted {
```
matzew: should we name them ...Event?  or move these events into its own package ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/jms/BatchLoadedProducer.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.jms.Queue;
+
+import org.jboss.aerogear.unifiedpush.message.TokenLoader;
+import org.jboss.aerogear.unifiedpush.message.holder.BatchLoaded;
+
+/**
+ * Receives message from {@link TokenLoader} and queues it under the variantID as a JMS message property in order to allow
+ * consumers simply receiving the message for given variant
+ */
+public class BatchLoadedProducer extends AbstractJMSMessageProducer {
```
matzew: not sure I get that class

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -118,37 +139,52 @@ public void sendPushMessage(final Variant variant, final Collection<String> toke
...
+            // trigger asynchronous deletion:
+            if (! transformedTokens.isEmpty()) {
+                logger.info("Deleting '" + inactiveTokens.size() + "' invalid iOS installations");
+                clientInstallationService.removeInstallationsForVariantByDeviceTokens(iOSVariant.getVariantID(), transformedTokens);
+            }
+            apnsServiceCache.queueFreedUpService(pushMessageInformationId, iOSVariant.getVariantID(), service);
+            callback.onSuccess();
+        } catch (Exception e) {
+            try {
+                service.stop();
```
matzew: is the stop here, in this class, correct ?

push/sender/src/test/java/org/jboss/aerogear/unifiedpush/message/AbstractJMSTest.java
```diff
@@ -0,0 +1,71 @@
+package org.jboss.aerogear.unifiedpush.message;
```
matzew: license

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
lfryc: Nice, I will use it.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/AllBatchesLoaded.java
lfryc: not exactly, when the loading process is done, it does not mean Senders processed those batches.

See `VariantCompleted` which distinguishes the end of whole sending process.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/exception/PushNetworkUnreachableException.java
lfryc: yea, it is a bit artificial atm, we need to really used that in Sender impls - the thing is that we need to distinguishing failure before we started pushing and after we started pushing... in a first case, we can allow to recover, in second case, it's no longer possible, since some tokens (perhaps all?) may be already pushed

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/BatchLoaded.java
lfryc: yes, this is for just one batch, allows to count them

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/holder/VariantCompleted.java
lfryc: +1 for package + renaming

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/jms/BatchLoadedProducer.java
lfryc: This is CDI->JMS bridge :-)

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
lfryc: nope, it is not correct, nice catch!

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java
```diff
@@ -46,9 +53,6 @@
 @Stateless
 public class TokenLoader {
 
-    private final static int NUMBER_OF_BATCHES = 10;
-    private final static int BATCH_SIZE = 1000;
```
matzew: These guys are still referenced on the JavaDoc: https://github.com/lfryc/aerogear-unifiedpush-server/blob/service-reuse-and-batch-size/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java#L80-L88

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfiguration.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+public class SenderConfiguration {
+
+    private int batchesToLoad;
+    private int batchSize;
+
+    public SenderConfiguration() {
+    }
+
+    public SenderConfiguration(int batchesToLoad, int batchSize) {
```
matzew: let's add some javadoc here

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Produces @ApplicationScoped @SenderType(VariantType.ADM)
+    public SenderConfiguration produceAdmConfiguration() {
+        return loadConfigurationFor(VariantType.ADM, new SenderConfiguration(10, 1000));
+    }
+
+    @Produces @ApplicationScoped @SenderType(VariantType.IOS)
+    public SenderConfiguration produceIosConfiguration() {
+        return loadConfigurationFor(VariantType.IOS, new SenderConfiguration(3, 10000));
```
matzew: I understand that we want 10k of tokens, per patch. but I don't understand the first argument.

matzew:
This call returns null: https://github.com/lfryc/aerogear-unifiedpush-server/blob/service-reuse-and-batch-size/service/src/main/java/org/jboss/aerogear/unifiedpush/service/metrics/PushMessageMetricsService.java#L162

and the server  has some long stack :-)

matzew:
I changed the token size to 1, and had two devices.
I noticed the distribution works fine

matzew:
a bit more of testing here

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
lfryc: That is the number of batches that will be loaded in one transaction. The transaction may not end on time, because the JMS queue will block it until it has enough memory to receive it.

So the number of batches loaded in one transaction is a trade-off between performance and need for repeating the work that was already done.

Details in:
https://docs.google.com/a/fryc.eu/drawings/d/13IsJWPSJNYXtst-UVxQYmzH36C_EXQMYYr_jcu7nFmE/edit?usp=drive_web

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/ApnsServiceCache.java
matzew: and add some comment, that we only want max. 10 concurrent connections

(we can increase the value later)

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
matzew: Mind adding some javadoc around this on the config. class ?

lfryc:
> This call returns null: https://github.com/lfryc/aerogear-unifiedpush-server/blob/service-reuse-and-batch-size/service/src/main/java/org/jboss/aerogear/unifiedpush/service/metrics/PushMessageMetricsService.java#L162

No longer, it generally helps to save the updates we do in `MetricsCollector` into database :-)

Moved the inactive tokens cleanup of `ApnsService` to the time when variant sending is completed and instantiated services are stopped.

Added few more debug log messages so that one can watch how cache works.

cvasilak:
registered two iOS devices and an Android variant, messages sent successfully :+1: 

Sending a message with ```-Daerogear.ios.batchSize=1`` configured can see in the console the split in batch:

```
18:04:57,450 INFO  [NotificationRouter] (default task-21) Processing send request with &apos;[alert=Bar, criteria=Criteria{aliases=null, deviceTypes=null, categories=null, variants=null}, time-to-live=Config{timeToLive=-1}]&apos; payload
18:04:57,720 INFO  [APNsPushNotificationSender] (Thread-5 (HornetQ-client-global-threads-672605525)) Sending payload for [1] devices to GCM
18:04:57,732 INFO  [ApnsServiceCache] (default task-21) Message submitted to PushNetworks for further processing
18:04:58,470 INFO  [GCMPushNotificationSender] (Thread-5 (HornetQ-client-global-threads-672605525)) Message to GCM has been submitted
18:04:59,160 INFO  [JPAPushMessageInformationDao] (Thread-9 (HornetQ-client-global-threads-672605525)) One batch to APNs has been submitted
18:04:59,277 INFO  [JPAPushMessageInformationDao] (Thread-1 (HornetQ-client-global-threads-672605525)) One batch to APNs has been submitted
```

@lfryc let me know if you need further testing.

iphone5s  iphone6 Android

matzew:
the NPE is gone! yay :tada:

matzew:
@lfryc I have two devices and running w/  `-Daerogear.ios.batchSize=1` shows me that there is only one receiver - looks like the counting is wrong (used to work).

Opening the app, using the push, on each device gives me afterwards the following on the UI:
```
1 receivers / 2 opened
```

matzew:
There is something, under load:

If I send a message every 100 ms, the UI shows: `0 receivers / 0 opened`

here is the sender code:
```
    public static void main(String... args) throws InterruptedException {


        PushSender defaultPushSender = DefaultPushSender.withRootServerURL("http://localhost:8080/ag-push")
                .pushApplicationId("fea4afad-9e4a-43a4-a8cc-e674ee82c979")
                .masterSecret("9e11c8d7-c2b3-4462-bb60-614b6538034d")
                .build();


        while (COUNTER < 100000) {

            final String payload = "This is only a test at " + new Date();

            final UnifiedMessage unifiedMessage = UnifiedMessage.withMessage()
                    .alert(payload)
                    .build();

            try {
                defaultPushSender.send(unifiedMessage);
            } catch (Exception e) {
                ;;
            }

            Thread.sleep(TimeUnit.MILLISECONDS.toMillis(100));

            COUNTER++;
        }


        try {
            Thread.currentThread().join();
        }
        catch (InterruptedException e) {

        }
    }
```

matzew:
@lfryc let's not worry about the 100ms sent frequenence, now - it's an unrealistic one.

The receiver incorrectness is more of a bug.

lfryc:
@matzew theoretically 100ms delay sending should be fixed as well

matzew:
-1 on `@Version `

lfryc:
forced pushed: update please, I've sent test fix

matzew:
:heart:

lfryc:
The last commit resolves issues with entities that are not written into database in `NotificationRouter` soon enough to be picked by at the end of the chain by `MetricsCollector`.

Still I'm getting the issue `EJBTransactionRolledbackException: entity not in the persistence context` [1]. I've verified it fails on both WF 8.2 and EAP 6.4, but currently do no understand why is that - if I obtain fresh entity from `EntityManager`, why it is not managed? :-)

But note that `EJBTransactionRolledbackException: entity not in the persistence context`  happens in 1-3 pushes from 1000. I would suggest open an separate issue and address there.

[1] https://gist.github.com/lfryc/6e67b747acc51b7dd97c

matzew:
is that having @Version?



On Tuesday, May 19, 2015, Lukáš Fryč <notifications@github.com> wrote:

> The last commit resolves issues with entities that are not written into
> database in NotificationRouter soon enough to be picked by at the end of
> the chain by MetricsCollector.
>
> Still I'm getting the issue EJBTransactionRolledbackException: entity not
> in the persistence context [1]. I've verified it fails on both WF 8.2 and
> EAP 6.4, but currently do no understand why is that - if I obtain fresh
> entity from EntityManager, why it is not managed? :-)
>
> But note that EJBTransactionRolledbackException: entity not in the
> persistence context happens in 1-3 pushes from 1000. I would suggest open
> an separate issue and address there.
>
> [1] https://gist.github.com/lfryc/6e67b747acc51b7dd97c
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/557#issuecomment-103378898>
> .
>


-- 
Sent from Gmail Mobile

lfryc:
Nope, this is without `@Version`

matzew:
perhaps @edewit can take a look? or gunnar?

On Tuesday, May 19, 2015, Lukáš Fryč <notifications@github.com> wrote:

> Nope, this is without @Version
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/557#issuecomment-103381927>
> .
>


-- 
Sent from Gmail Mobile

lfryc:
Ok, I believe I found the issue, the `NotificationRouter` was sending the messages for dispatching events in a non-transacted way, which in turns mean that in fact the underlying entity does not have to be persisted before it goes propagates the rest of the messaging chain.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/MetricsCollector.java
```diff
@@ -16,40 +16,69 @@
...
+
     /**
      * Receives metrics and update the push message information in database.
      *
      * @param variantMetricInformation the variant metrics info object
      */
     public void collectMetrics(@Observes @Dequeue VariantMetricInformation variantMetricInformation) {
-        PushMessageInformation pushMessageInformation = variantMetricInformation.getPushMessageInformation();
+        PushMessageInformation pushMessageInformation = metricsService.getPushMessageInformation(variantMetricInformation.getPushMessageInformation().getId());
+        metricsService.lock(pushMessageInformation);
```
edewit: Do we really need to lock by hand? Isn't Optimistic locking on by default?

lfryc: Nope, we had concurrency issues here - if you send really small transactions or send them quickly, lot of `VariantMetricInformation` are coming concurrently.

Two `MetricsCollector`s were able to work concurrently on one `PushMessageInformation`, thus rewriting their updates.

We could do optimistic locking with `@Version` though, should I open a JIRA? @matzew didn't want to introduce that now, but maybe later?

If lock fails / times out, the transaction is aborted and `VariantMetricInformation` is delivered later.

lfryc:
Introducing XA transactions got rid of the `entity not in the persistence context` issue. I'm sending 1000 messages in a row without any issue on WF 8.2 with MySQL and EAP 6.4 with PoistgreSQL.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
lfryc: It's already there, but I will clarify and comment on a contructor.

matzew:
@lfryc perfect

lfryc:
everything should be done now @matzew

lfryc:
rebasing...

lfryc:
rebased + everything should be fixed as per PR comments ^ /cc @matzew

matzew:
landed in 9feff40e94b0bfd40e5c5c1fe5a1c1ff8a5935fb

Thanks @lfryc :tada:

PR #560, opened 2015-05-18
AGPUSH-1404 : change the warning label

sebastienblanc:
@lfryc wdyt of this ?

### DIFF
--- a/admin-ui/app/index.html
+++ b/admin-ui/app/index.html
@@ -98,7 +98,7 @@
                   <span class="pficon pficon-warning-triangle"></span>
                   <span class="pficon pficon-warning-exclamation"></span>
               </span>
-              <a href ng-link="appDetail({app: warning.applicationID, tab: 'activity'})"><strong>Push failed:</strong> {{ warning.applicationName }}</a>
+              <a href ng-link="appDetail({app: warning.applicationID, tab: 'activity'})">Push Problem for Variant <strong>{{ warning.variant.name }}</strong></a>
             </li>
           </ul>
           <!--<div class="footer"><a href="#">Clear Warnings</a></div>-->


admin-ui/app/index.html
```diff
@@ -98,7 +98,7 @@
                   <span class="pficon pficon-warning-triangle"></span>
                   <span class="pficon pficon-warning-exclamation"></span>
               </span>
-              <a href ng-link="appDetail({app: warning.applicationID, tab: 'activity'})"><strong>Push failed:</strong> {{ warning.applicationName }}</a>
+              <a href ng-link="appDetail({app: warning.applicationID, tab: 'activity'})">Variant <strong>{{ warning.variant.name }}</strong> has push failures.</a>
```
matzew: failure? or failures?

sebastienblanc: I can not tell from there if there on or more errors. I could write `failure(s)` ?

matzew: perhaps @lfryc or @edewit know a more fancy way :-)

lfryc: definitely, there is a `ups-pluralize` directive, or standard angular pluralize may work better here

sebastienblanc: Yeah that is the problem , at this point we don't know if there one or more failure and it would not be optimal to made another call just to retrieve that. 
Let's try to find a generic warning label.

matzew: -1 on an extra call

@lfryc @sebastienblanc 

How about:
```
Push Problem for Variant <strong>{{ warning.variant.name }}</strong>.
```

sebastienblanc:
updated with @matzew  suggestion

matzew:
I am fine w/ that.

@lfryc agree?

lfryc:
Works fine, merging..

lfryc:
Landed: https://github.com/aerogear/aerogear-unifiedpush-server/commit/280b7de8e28c1b79e66cb12db1b2d73e845ccf25

PR #566, opened 2015-05-21
Removing iteration for unneeded token spit

matzew:
done for https://issues.jboss.org/browse/AGPUSH-1419
@lfryc clean up of unneeded code

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfiguration.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfiguration.java
@@ -59,6 +59,10 @@ public class SenderConfiguration {
         return batchesToLoad;
     }
 
+    void setBatchesToLoad(int batchesToLoad) {
+        this.batchesToLoad = batchesToLoad;
+    }
+
     /**
      * Batch size specifies how many devices will be loaded and delivered in one batch.
      *
@@ -73,6 +77,10 @@ public class SenderConfiguration {
      */
     public int batchSize() {
         return batchSize;
+    }
+
+    void setBatchSize(int batchSize) {
+        this.batchSize = batchSize;
     }
 
     /**
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java
@@ -22,6 +22,7 @@ import javax.enterprise.inject.Produces;
 import org.jboss.aerogear.unifiedpush.api.VariantType;
 import org.jboss.aerogear.unifiedpush.message.sender.SenderType;
 import org.jboss.aerogear.unifiedpush.message.util.ConfigurationUtils;
+import org.jboss.aerogear.unifiedpush.utils.AeroGearLogger;
 
 /**
  * Loads and stores configuration for specific Push Networks.
@@ -40,6 +41,8 @@ import org.jboss.aerogear.unifiedpush.message.util.ConfigurationUtils;
  * @see SenderConfiguration
  */
 public class SenderConfigurationProvider {
+
+    private final AeroGearLogger logger = AeroGearLogger.getInstance(SenderConfigurationProvider.class);
 
     @Produces @ApplicationScoped @SenderType(VariantType.ANDROID)
     public SenderConfiguration produceAndroidConfiguration() {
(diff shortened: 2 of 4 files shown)


lfryc:
@matzew I guess we should enforce batches less than 1000 tokens in android configuration. Worth a JIRA probably

edewit:
Instead of waiting for the list to be empty and removing the elements it could iterate over the chunks:
```java
        for (int i = 0; i < registrationIDs.size() / GCM_PAGE + 1; i++) {
            int chunkSize = Math.min(registrationIDs.size(), (i + 1) * GCM_PAGE);
            List<String> sublist = registrationIDs.subList(i  * GCM_PAGE, chunkSize);
        }
```

matzew:
but we just send 1k tokenz to a GCMSender now. more does not make sense,
therefore the code remove

On Thursday, May 21, 2015, Erik Jan de Wit <notifications@github.com> wrote:

> Instead of waiting for the list to be empty and removing the elements it
> could iterate over the chunks:
>
>         for (int i = 0; i < registrationIDs.size() / GCM_PAGE + 1; i++) {
>             int chunkSize = Math.min(registrationIDs.size(), (i + 1) * GCM_PAGE);
>             List<String> sublist = registrationIDs.subList(i  * GCM_PAGE, chunkSize);
>         }
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/566#issuecomment-104341637>
> .
>


-- 
Sent from Gmail Mobile

delalis:
Hey guys, is this a specific fix that will address my concurrentModificationException?  

If so, when would you expect a build would be available for us to download?  

Thanks for addressing this so quickly.

fyi, here are the links I'm referring to:

https://issues.jboss.org/browse/AGPUSH-1419

http://aerogear-dev.1069024.n5.nabble.com/Android-GCM-PUSH-ERROR-Error-sending-payload-to-GCM-server-java-util-ConcurrentModificationException-td11655.html#a11664

matzew:
@delalis we are releasing 1.1.0-beta.1 early nexxt week.

your fix is for -beta.2, a few weeks later

edewit:
@matzew just out of curiosity where is this 1k token split now?

matzew:
https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java#L46

On Fri, May 22, 2015 at 9:33 AM, Erik Jan de Wit <notifications@github.com>
wrote:

> @matzew <https://github.com/matzew> just out of curiosity where is this
> 1k token split now?
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/566#issuecomment-104549940>
> .
>



-- 
Matthias Wessendorf

lfryc:
@edewit the configuration is than respected by the TokenLoader: https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java#L117

lfryc:
@matzew thinking about it a bit more, it still can be beneficial to do splitting to 1000 tokens in the sender itself:

say we want to optimize the flow, and we want to submit 5000 tokens in one batch for GCM sender, so that the sender will do 5 subsequent connections with 1000 devices within one transaction.

----

However in this state, if user reconfigures `-Daerogear.android.batchSize=1500`, the sender will fail, right?

matzew:
a few comments:
* So we do some logic to dispatch 5k tokens, in one batch.
* we add somewhat similar logic to split that to 1k per sender.

Wouldn't it be nice to have the sender not worrying about that ? 

I think the `-Daerogear.android.batchSize=1500` is not a good config. it should be ignored, until we have the XMPP based stuff

lfryc:
Okay, I'm fine with either of solutions, i will create an issue to cover
that.

matzew:
@lfryc what would be the actual benefit if we keep the iteration inside of the sender?
Less of resources?

sebastienblanc:
@matzew @lfryc are we in agreement for this PR ?

matzew:
Not sure :-)

lfryc:
If we keep sender splitting the GCM, we will make sure the impl is correct without being tied to the configuration of the component (TokenLoader) before.

Otherwise we need to make sure user won't be able to misconfigure (sanitize the case of configuration). We don't have any config validation ATM, but since we should also warn him he misconfigured in case of iOS batches or number of parallel iOS senders, we need to introduce this config validation anyway.

lfryc:
I.e. if we keep this impl, we have to sanitize the user configuration in https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/configuration/SenderConfigurationProvider.java#L46

lfryc:
Okay, I will send a PR here to do required configuration validation changes.

matzew:
@lfryc perfect!

lfryc:
Here you go: https://github.com/matzew/aerogear-unifiedpush-server/pull/8

PR #581, opened 2015-06-02
updated UI screenshot in README

lfryc:
The screenshot is not yet on the specified address, but it will be after merging,

you can preview how it looks like on this link: https://github.com/lfryc/aerogear-unifiedpush-server/tree/screenshot3


matzew:
landed, thanks

PR #597, opened 2015-06-18
change default Nothing Selected to All Variants on Send Notification dialog AGPUSH-1449

lfryc:
http://issues.jboss.org/browse/AGPUSH-1449

The Send Push Notification dialog should say "All Variants" by default and when checked some variants and then uncheck them, it should still say "All Variants".


matzew:
thanks!

PR #601, opened 2015-06-22
latest of KC

matzew:
@sebastienblanc please test and review

Btw. do you have a simple (Node.js) script for the direct access token grant ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
         <groupId>org.jboss.aerogear</groupId>
         <artifactId>aerogear-parent</artifactId>
         <!-- NOTE: if updated for version of UPS dependencies, please update the 'aerogear.bom.version' as well -->
-        <version>0.2.15</version>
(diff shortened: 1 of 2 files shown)


sebastienblanc:
@matzew I will update https://github.com/aerogear/aerogear-unifiedpush-server/blob/AGPUSH-1149_enable_direct_access_grant/directgrant.js dince I'm not sure it's still working with kc 1.3.1 and let you know

matzew:
@sebastienblanc not sure we need to really maintain the branch, perhaps we just check in a few of these scripts on a different folder?

just wondering, why should they no longer work ?

sebastienblanc:
:+1: 
All good , even  https://github.com/aerogear/aerogear-unifiedpush-server/blob/AGPUSH-1149_enable_direct_access_grant/directgrant.js is still working (just have to create a new KC client called `ups-client` , nothing to configure, the defailts settings are good)

PR #603, opened 2015-06-24
Using Skip instead of Cancel

matzew:
@andresgalante @lfryc Most of the screens use Skip, therefore doing the same for these two pages

### DIFF
--- a/admin-ui/app/components/welcome/welcome.html
+++ b/admin-ui/app/components/welcome/welcome.html
@@ -48,7 +48,7 @@
         </a>
       </p>
       <p ng-if="app.stats.applications > 0">
-        <a ng-link="home">Cancel the wizard</a>
+        <a ng-link="home">Skip the wizard</a>
       </p>
     </div><!-- col -->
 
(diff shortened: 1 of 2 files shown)


danielpassos:
:+1:

sebastienblanc:
+1 merging

andresgalante:
@matzew @sebastienblanc @danielpassos @lfryc "Cancel" on welcome and step1 was correct. The action is "cancel" creating a new app. 
The rest of the steps are skip, because the user would be skipping the wizard.

I recommend going back with the merge.

matzew:
@andresgalante ok - so 'Cancel' on both files (see changed files)

andresgalante:
@matzew yeap, the action is this:
The user clicks on "crate new app"
welcome screen: Gets instructions, he can either cancel or continue
step one: gets the option to name and create the app, again he can either cancel or continue

After this the app has already been crated, so he can no longer cancel it, he can just skip the wizard. Does it make sense?

PR #608, opened 2015-06-26
Speeding up, by adding a quick up-front validaiton

matzew:
@sebastienblanc please review

### DIFF
--- a/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/validation/DeviceTokenValidator.java
+++ b/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/validation/DeviceTokenValidator.java
@@ -68,6 +68,18 @@ public class DeviceTokenValidator implements ConstraintValidator<DeviceTokenChec
         }
         final VariantType type = installation.getVariant().getType();
 
+        return isValidDeviceTokenForVariant(deviceToken, type);
+    }
+
+
+    /**
+     * Helper to run quick up-front validations.
+     *
+     * @param deviceToken the submitted device token
+     * @param type type of the variant
+     * @return true if the token is valid
+     */
+    public static boolean isValidDeviceTokenForVariant(final String deviceToken, final VariantType type) {
         switch (type) {
             case IOS:
                 return IOS_DEVICE_TOKEN.matcher(deviceToken).matches();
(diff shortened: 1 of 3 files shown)


model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/validation/DeviceTokenValidator.java
```diff
@@ -68,6 +68,18 @@ public boolean isValid(Installation installation, ConstraintValidatorContext con
...
+
+
+    /**
+     * Helper to run quick up-front validations.
+     *
+     * @param deviceToken the submitted device token
+     * @param type type of the variant
+     * @return true if the token is valid
+     */
+    public static boolean isValidDeviceTokenForVariant(final String deviceToken, final VariantType type) {
```
sebastienblanc: Would be nice to have an Unit test for this static method.

sebastienblanc: Maybe for another PR but I think we should log when the token is invalid, now the only feedback , is the endpoint returning a bad request but it can be hard for one to find out the reason.

sebastienblanc:
appart from my comments :+1:

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/validation/DeviceTokenValidator.java
matzew: >  Would be nice to have an Unit test for this static method.

Sure! Added simple test (there are already tests for the internal code, on the JPA layer)

matzew: > Maybe for another PR but I think we should log when the token is invalid, now the only feedback , is the endpoint returning a bad request but it can be hard for one to find out the reason.

Ok, I added a finest logger, since I would like to have as miminmal noice on the console. When a real device sends an invalid token - there is usally not much to see on the console for the guy running the device.

For development, well, yeah - but our API contract stats:
```
400 - (Bad Request) The format of the client request was incorrect (e.g. missing required values).
```

matzew: @sebastienblanc I think... we could update the API doc to say that the deice token may be also invalid ?

PR #613, opened 2015-06-29
Shows more information on variants

andresgalante:
@matzew @sebastienblanc I've added a structure to put more information for each variant.

**To Do:** add the correct Angular variables to show the correct information.

It looks like this:
img

What do you think?


matzew:
:eyeglasses: at ti

PR #615, opened 2015-07-03
making 'developer' default role for new users

matzew:
For convenience, when creating new users, the default role should be our developer role!


sebastienblanc:
:+1:

PR #619, opened 2015-07-21
Bootstrap view

lfryc:
@matzew @sebastienblanc could you please review this and test it? :eyes: 

*testing:*

You can access the bootstrap view on http://localhost:8080/ag-push/#/bootstrap
Once you hit Enable Push, you will be redirected to the app detail and you will see list of variants.

Push Application name is generated and Variant names are copy of their VariantType (android, windows_wns, etc.)

### DIFF
--- a/admin-ui/app/components/bootstrap/bootstrap.html
+++ b/admin-ui/app/components/bootstrap/bootstrap.html
@@ -17,16 +17,16 @@
 
   <div ng-if="bootstrap.types.android">
     <div class="form-group">
-      <label class="col-sm-3 control-label" for="gcmApiKey">Push Network</label>
+      <label class="col-sm-3 control-label" for="gcmApiKey"></label>
       <div class="col-sm-7"><em>Google Cloud Messaging</em> Key<br>
-        <input type="text" placeholder="e.g. 5a448c2f31700a466fbc9d33d33942b043a27596" id="gcmApiKey" class="form-control" ng-model="variant.googleKey" ng-required="variant.type == 'android'">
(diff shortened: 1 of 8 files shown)


matzew:
@lfryc awesome stuff, dude!

I wonder if we should actually enable it in 1.2.0 - or even 1.1.1 ? 

PErhaps @andresgalante has some ideas on integrating this view in to the regular UI ?

lfryc:
@andresgalante what do you think? does it fit into the current UI?


@kpiwko this hidden view could be interesting for you guys as well:
http://localhost:8080/ag-push/#/bootstrap

it is a way faster to start with for experienced users than the wizard :-)

andresgalante:
@lfryc @matzew 
I can think on a way to include it, I am just wondering how useful for the user it can be. Do you think we have a use case for a user creating more than one variant at the same time?

matzew:
I guess not now - it's perhaps really more for API based hooks.

LEts keep as is, I guess ... :-)

If we ever have to... we now have a hidden UI :-)

On Wed, Jul 22, 2015 at 4:38 PM, Andres Galante <notifications@github.com>
wrote:

> @lfryc <https://github.com/lfryc> @matzew <https://github.com/matzew>
> I can think on a way to include it, I am just wondering how useful for the
> user it can be. Do you think we have a use case for a user creating more
> than one variant at the same time?
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/619#issuecomment-123744362>
> .
>



-- 
Matthias Wessendorf

kpiwko:
@lfryc nice!

PR #624, opened 2015-08-04
Enabling setting the finalName of the dist build with a property

secondsun:
This lets us name the zip from dist from the command line.

To test :
 * mvn clean install
 * cd dist
 * mvn clean install -DfinalName=unifiedpush
 * check that there is a unifiedpush-dist.zip & tgz file in target


danbev:
:+1: 
Just noticed that the system property is missing the ```=``` sign in the test description. Should be:
```
$ mvn clean install -DfinalName=unifiedpush
```

secondsun:
Sebi pulled this into his pr, closing.

PR #626, opened 2015-08-06
Gcm topics

secondsun:
This is a bit of a big one so it needs lots of eyes and lots of commentary.  I hope.

 # How to test :
* Build the project. 
* Build the https://github.com/aerogear/aerogear-android-push/pull/50 for AGDROID
  *
* Configure a push variant etc etc in UPS
* Configure a sample application (I used the hello world)
* In the sample application add a category "topic" (or anything you want) to the Push Registrar config
* In UPS sent a message to the device. In the UPS log you will see Sending payload for [1] devices to GCM
* Now send a message but set the category to "topic" (or whatever you set in step *)
* In UPS you should see Sending payload for [1] topics to GCM
* You should also see both messages in the app.

PR #630, opened 2015-08-12
Gcm topics

secondsun:
This is a migration of the PR from https://github.com/aerogear/aerogear-unifiedpush-server/pull/626

### DIFF
--- a/gcm/CONTRIBUTING.md
+++ b/gcm/CONTRIBUTING.md
@@ -1,0 +1,36 @@
+# How to become a contributor and submit your own code
+
+## Contributor License Agreements
+
+We'd love to accept your sample apps and patches! Before we can take them, we
+have to jump a couple of legal hurdles.
+
+Please fill out either the individual or corporate Contributor License Agreement
+(CLA).
+
+  * If you are an individual writing original source code and you're sure you
+    own the intellectual property, then you'll need to sign an [individual CLA]
+    (https://cla.developers.google.com/about/google-individual).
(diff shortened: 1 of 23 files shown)


matzew:
the gcm/ folder... should be just temporary - right?
IMO best is: we create a XMPP client lib (standalone) and use it as a dependency inside the UPS/GCM component

secondsun:
@matzew I don't think the XMPP stuff is going to happen for a long time.  There is still a lot of work to be done on topics that I didn't expect (specifically how do we support alias and devicetype).

matzew:
thats fine. But sounds reasonable to aim for 2.0 having XMPP?

the throughput via XMPP is also better, so worth to plan for it, in the
future ;-)

On Wednesday, August 12, 2015, Hoyt Summers Pittman <
notifications@github.com> wrote:

> @matzew <https://github.com/matzew> I don't think the XMPP stuff is going
> to happen for a long time. There is still a lot of work to be done on
> topics that I didn't expect (specifically how do we support alias and
> devicetype).
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/630#issuecomment-130397506>
> .
>


-- 
Sent from Gmail Mobile

secondsun:
@matzew I think 2.0 it is reasonable

matzew:
there will be a newer PR for this work

matzew:
Test run #1: https://github.com/aerogear/aerogear-unifiedpush-server/pull/708

PR #633, opened 2015-09-08
Fixed MireDot issues

idelpivnitskiy:
* added missing `@Produces` annotations
* added comment for `search` param at InstallationManagementEndpoint.findInstallations(...)
* removed snapshots repository

Result: http://condor.pp.ua/aerogear-ups-rest-api-doc/fix-miredot-issues/

Note: we also could update the version of miredot. The latest release wasn't be major: http://miredot.com/docs/release-notes/
But I'm not sure that it is a great idea before the final release. WDYT?

### DIFF
--- a/jaxrs/pom.xml
+++ b/jaxrs/pom.xml
@@ -242,14 +242,6 @@
             <id>miredot</id>
             <name>MireDot Releases</name>
             <url>http://nexus.qmino.com/content/repositories/miredot</url>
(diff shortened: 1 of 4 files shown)


lfryc:
:eyes: looking

matzew:
@lfryc ready to merge ?

lfryc:
yes, ready :-)

PR #635, opened 2015-10-01
AGPUSH-1512: workaround: recount receivers in UPS Console

lfryc:
@sebastienblanc could you please test this?

This is workaround that should do a math of serverVariants/totalVariants/receivers on client.

https://issues.jboss.org/browse/AGPUSH-1512


lfryc:
This is not correct, it shows request as completed when 0 batches were served

PR #636, opened 2015-10-01
AGPUSH-1513: (workaround) recount receivers in UPS Console

lfryc:
@sebastienblanc could you please test this?

This is workaround that should do a math of serverVariants/totalVariants/receivers on client.

https://issues.jboss.org/browse/AGPUSH-1513


lfryc:
forced pushed another commit, the commit message was referencing different issue number, corrected now

PR #639, opened 2015-10-08
Use httpcomponent properties declared in AeroGear BOM (no overrides)

vibe13:
To be merged after https://github.com/aerogear/aerogear-parent/pull/52

### DIFF
--- a/servers/auth-server/pom.xml
+++ b/servers/auth-server/pom.xml
@@ -71,13 +71,11 @@
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <version>4.3.6</version>
         </dependency>
 
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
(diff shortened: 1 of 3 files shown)


matzew:
will there be no code change required?

lfryc:
as long as the build passes I guess we are good to go

vibe13:
@matzew @lfryc yes exactly, this PR uses the same versions currently in use, but instead of declaring unused versions in aerogear-parent and overriding them in UPS, it changes the versions in aerogear-parent with the ones currently used, and gets rid of the override in UPS.

matzew:
@lfryc @vibe13 should we include this in 1.1.0 final too ?

vibe13:
@matzew hi Matt this is more a nice to have, but not essential for the release. It would require a new release of aerogear-parent also (after https://github.com/aerogear/aerogear-parent/pull/52 merge), and the inclusion if this new parent

matzew:
ok, thanks - I missed that cross reference. not merging :)

PR #642, opened 2015-11-19
allow to run migrate script from source AGPUSH-1527

lfryc:
https://issues.jboss.org/browse/AGPUSH-1527

### DIFF
--- a/migrator/README.md
+++ b/migrator/README.md
@@ -52,7 +52,15 @@ In case of a successful run, the script prints
 If you want to get some more details on the actual migration steps, enable logging while performing the database migration:
 
     ./bin/ups-migrator --logLevel=DEBUG update
+    
+#### Running the script from source
 
+If you are developing UnifiedPush Server, you can run the script from source instead of from distribution:
+
+    cd aerogear-unifiedpush-server/migrator/
+    mvn clean install dependency:copy-dependencies -DincludeScope=runtime
+    export UPS_MIGRATOR_HOME=$PWD/target/dependency/*:$PWD/target
+    sh ./src/main/shell/ups-migrator update
 
 ### WAR migration
 
--- a/migrator/pom.xml
+++ b/migrator/pom.xml
@@ -33,6 +33,7 @@
     </dependencies>
 
     <build>
+        <finalName>ups-migrator</finalName>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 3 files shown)


lfryc:
@matzew @danbev could you please review?

migrator/README.md
```diff
@@ -52,7 +52,16 @@ In case of a successful run, the script prints
...
 
     ./bin/ups-migrator --logLevel=DEBUG update
+    
+#### Running the script from source
 
+If you are developing UnifiedPush Server, you can run the script from source instead of from distribution:
+
+    cd aerogear-unifiedpush-server/migrator/
+    mvn clean install
+    mvn dependency:copy-dependencies -DincludeScope=runtime
```
lfryc: should we make it `mvn clean install dependency:copy-dependencies -DincludeScope=runtime` instead?

matzew:
the change from `644` to `755` on:
* migrator/src/main/shell/ups-migrator 

not sure, the source code should be executable, tbh

lfryc:
@matzew right, I had `sh ./src/main/shell/ups-migrator update` first, but it is platform dependent.. if we forget about windows, then we can do that...

matzew:
yep - windows - we need a BAT. Sebi has a ticket for that, for future or
something like that

On Thu, Nov 19, 2015 at 12:17 PM, Lukáš Fryč <notifications@github.com>
wrote:

> @matzew <https://github.com/matzew> right, I had sh
> ./src/main/shell/ups-migrator update first, but it is platform
> dependent.. if we forget about windows, then we can do that...
>
> —
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/642#issuecomment-158026380>
> .
>



-- 
Matthias Wessendorf

migrator/README.md
danbev: Would it be worth having a a profile that does this, like ```mvn -PsetupEnv``` or something like that?

danbev:
:+1: Works like advertised.

lfryc:
Thanks for testing @danbev ,

I've incorporated README change suggested by @matzew and merged mvn to one-liner..

matzew:
@lfryc I still see the 644 -> 755 change on the migrator source file - should we really do that ?

lfryc:
Yep, sry, nice catch, lemme fix that

lfryc:
@matzew should be fine now

matzew:
:+1:

PR #647, opened 2015-11-30
Improve Developer experience

matzew:
Adding persistence.xml properties for different DDL values, based on needs. For development (default) the `persistence.xml` will use `update` and for production the file will contain `validate`.

With this change it is no longer needed to run the migrator when deploying the latest from master.

However, for our production profiles (`openshift` and `dist`) the `validate` value will be injected, since in this environment we do need to make sure we have proper migration tool execution

### DIFF
--- a/model/jpa/pom.xml
+++ b/model/jpa/pom.xml
@@ -27,7 +27,7 @@
     <name>UnifiedPush Server Model JPA implementation</name>
 
     <dependencies>
-		
+        
         <dependency>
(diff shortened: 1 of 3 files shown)


matzew:
@sebastienblanc @lfryc @abstractj mind giving this a shot ?

lfryc:
:+1: nice stuff, merging

lfryc:
:rocket: landed...

PR #651, opened 2015-12-07
Adding details about the actual variant on PUT request for update

matzew:
@wei-lee @sebastienblanc please review

### DIFF
--- a/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
+++ b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
@@ -138,7 +138,7 @@ public class WindowsVariantEndpoint extends AbstractVariantEndpoint {
             windowsVariant.setName(updatedWindowsVariant.getName());
             windowsVariant.setDescription(updatedWindowsVariant.getDescription());
             variantService.updateVariant(windowsVariant);
-            return Response.noContent().build();
+            return Response.ok(windowsVariant).build();
         }
 
         return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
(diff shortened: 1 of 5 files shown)


lfryc:
code-wise seems good, all 5 supported platforms covered

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
```diff
@@ -138,7 +138,7 @@ public Response updateWindowsVariation(
             windowsVariant.setName(updatedWindowsVariant.getName());
             windowsVariant.setDescription(updatedWindowsVariant.getDescription());
             variantService.updateVariant(windowsVariant);
-            return Response.noContent().build();
+            return Response.ok(windowsVariant).build();
```
sebastienblanc: I wonder if must not retrieve the correct subtype like we do in the `GET` , https://github.com/matzew/aerogear-unifiedpush-server/blob/ReturnVariantDetailsOnUpdate/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java#L96

lfryc: There we list all Windows variants, while here the variant is known.

lfryc:
The build fails for unrelated reason, could you someone hit Rebuild button?

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
sebastienblanc: Ah right :) , I must wake up :)

matzew: is the update (one) variant endpoint :)

matzew:
@lfryc it crashes in the migrator - I have no clue :) 

totally unrelated

wei-lee:
@matzew Thanks for the update!

PR #648, opened 2015-12-01
RFC - add support for node 4.2.2

lholmquist:
For https://issues.jboss.org/browse/AGPUSH-1541,  i needed to update the grunt-uglify task for everything to work.

We should also update the other dependecies to the latest.

Perhaps in the same PR.   I also added node 0.12 and 4.2.2 to travis, so we will see if everthting passes first  :).  also removed 0.8 from travis

PR #652, opened 2015-12-07
adding bower components to the ignore

matzew:
Looks like left over from https://github.com/aerogear/aerogear-unifiedpush-server/pull/648

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ release.properties
 
 .DS_Store
 admin-ui/node
+admin-ui/app/bower-components/


matzew:
@lfryc @lholmquist is this OK ?

lfryc:
@matzew interesting, I would guess it should be ignored by the `.gitignore` specified in `admin-ui` subfolder: https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/admin-ui/.gitignore#L6

matzew:
ok - will update, @lfryc 

btw. same build failure

matzew:
see https://github.com/aerogear/aerogear-unifiedpush-server/pull/655

PR #653, opened 2015-12-07
Adding check for more queries and NPE fix for messed up DB

matzew:
Adding a NPE check in case the PA is not in the DB - therefore, the .getName() blows it


sebastienblanc:
:+1:

PR #656, opened 2015-12-09
providing slightly better GCM placeholder value

matzew:
@sebastienblanc here we go, I applied your suggestion from AGPUSH-1540


sebastienblanc:
:+1:

PR #674, opened 2016-01-25
AGPUSH-1567 : fix the ttl computing for apns

sebastienblanc:


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java
```diff
@@ -208,10 +208,10 @@ private Date createFutureDateBasedOnTTL(int ttl) {
 
         // no TTL was specified on the payload, we use the MAX Default from the APNs library:
         if (ttl == -1) {
-            return new Date(System.currentTimeMillis() + EnhancedApnsNotification.MAXIMUM_EXPIRY * 1000L);
+            return new Date(EnhancedApnsNotification.MAXIMUM_EXPIRY * 1000L);
```
edewit: `EnhancedApnsNotification.MAXIMUM_EXPIRY` is  `Integer.MAX_VALUE` [ref](https://github.com/notnoop/java-apns/blob/598b396e93221ffb7373c95ec760c2fec07601a9/src/main/java/com/notnoop/apns/EnhancedApnsNotification.java#L57) maybe we should use: `new Date(Long.MAX_VALUE)` that way it's the maximum date value?

lfryc: Hmm, I'd prefer to use the project specific reference as @sebastienblanc did, since that can change in the future.

edewit: @lfryc `EnhancedApnsNotification.MAXIMUM_EXPIRY` is just a constant it's not meant to be turned into a date. If we remove `System.currentTimeMillis()` the result will be: `Sun Jan 25 21:31:23 CET 1970` so either we leave it as it was or we change it to `new Date(Long.MAX_VALUE)` both options are fine with me, but the change as it it proposed here is wrong

lfryc: Ah, noticed the slight difference here, make sense @edewit

sebastienblanc: I'm confused now, what needs to be changed then ?

sebastienblanc:
landed https://github.com/aerogear/aerogear-unifiedpush-server/commit/7cdb15f8b9606f2e5ec677ab978d02c355e39620

PR #687, opened 2016-03-10
Update mergeCategories

matzew:
Done for AGPUSH-1593

removes non persistent categories and replaces with their matching counterparts from the DB. afterwards the new ones are just set/applied

### DIFF
--- a/service/src/test/java/org/jboss/aerogear/unifiedpush/service/ClientInstallationServiceTest.java
+++ b/service/src/test/java/org/jboss/aerogear/unifiedpush/service/ClientInstallationServiceTest.java
@@ -20,13 +20,11 @@ import org.jboss.aerogear.unifiedpush.api.AndroidVariant;
 import org.jboss.aerogear.unifiedpush.api.Category;
 import org.jboss.aerogear.unifiedpush.api.Installation;
 import org.jboss.aerogear.unifiedpush.api.iOSVariant;
-import org.jboss.aerogear.unifiedpush.dao.InstallationDao;
 import org.jboss.aerogear.unifiedpush.dao.ResultStreamException;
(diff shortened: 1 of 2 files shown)


matzew:
@edewit mind taking a look here ?

edewit:
:eyes:

service/src/test/java/org/jboss/aerogear/unifiedpush/service/ClientInstallationServiceTest.java
```diff
@@ -101,6 +96,100 @@ public void registerDevices() {
     }
 
     @Test
+    public void registerDevicesWithCategories() {
+
+        Installation device = new Installation();
+        String deviceToken = generateFakedDeviceTokenString().toUpperCase();
+        device.setDeviceToken(deviceToken);
+        final Set<Category> categories = new HashSet<Category>(Arrays.asList(new Category("football"), new Category("football")));
```
edewit: This will result in one category, because it's a Set, do we really want to write it like that?

edewit:
Much simpler and it works :+1:

service/src/test/java/org/jboss/aerogear/unifiedpush/service/ClientInstallationServiceTest.java
matzew: yeah, not sure :smile: 

I think the motivation was to simulate a device that (accidentally?) registers twice (or more often) the same category, in one request. B

But yeah, since a SET, of course :) there is only one. Happy to improve the test

matzew:
> Much simpler and it works :+1:

Thanks a lot for testing.  Not sure, if that's really the ultimate change. But, indeed, it works :smile:

PR #691, opened 2016-03-15
Adding GCM token validation with Canonical ID 

sinarz:
From GCM docs:

> When you send a notification to a registration id which is expired, for the 1st time the message(notification) will be delivered **but you will get a new registration id with the name canonical id. Which mean, the registration id you sent the message to has been changed to this canonical id, so change it on your server side as well.**
[https://developers.google.com/cloud-messaging/registration](https://developers.google.com/cloud-messaging/registration)

> If a bug in the client app triggers multiple registrations for the same device, it can be hard to reconcile state and the client app might end up with duplicate messages.
Implementing canonical IDs can help you more easily recover from these situations. A canonical registration ID is the registration token of the last registration requested by the client app . This is the ID that the server should use when sending messages to the device.
If you try to send a message using an old registration token, GCM will process the request as usual, but it will include the canonical ID in the registration_id field of the response. **Make sure to replace the registration token stored in your server with this canonical ID, as eventually the old registration token will stop working.**

I submitted a GCM token in aerogear and used it for some time. after that, the token was expired, and it was removed from aerogear database and not replaced with given canonical id.

I checked aerogear code and the feature was not implemented in the sender file, so i updated it:
`org.jboss.aerogear.unifiedpush.message.sender.GCMPushNotificationSender.java`

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/GCMPushNotificationSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/GCMPushNotificationSender.java
@@ -171,17 +171,25 @@ public class GCMPushNotificationSender implements PushNotificationSender {
             //check if current index of result has canonical id
             String canonicalRegId = result.getCanonicalRegistrationId();
             if (canonicalRegId != null) {
-                // same device has more than one registration id: update it
-              
-                // find the installation and change its token to new canonical id
-                Installation installation = clientInstallationService.findInstallationForVariantByDeviceToken(variantID,registrationIDs.get(i));
-                installation.setDeviceToken(canonicalRegId);
-              
-                //update installation with the new token
-                logger.info(String.format("Based on returned canonical id from GCM, updating Android installations with registration id [%s] with new token [%s] ", registrationIDs.get(i), canonicalRegId));
-                clientInstallationService.updateInstallation(installation);
-              
-                //if the result contains a canonical id, there is no need to remove the current inactive token from data store
+                // same device has more than one registration id: update it, if needed!
+                // let's see if the canonical id is already in our system:
+                Installation installation = clientInstallationService.findInstallationForVariantByDeviceToken(variantID, canonicalRegId);
+
+                if (installation != null) {
+                    // ok, there is already a device, with newest/latest registration ID (aka canonical id)
+                    // It is time to remove the old reg id, to avoid duplicated messages in the future!
+                    inactiveTokens.add(registrationIDs.get(i));
+
+                } else {
+                    // since there is no registered device with newest/latest registration ID (aka canonical id),
+                    // this means the new token/regId was never stored on the server. Let's update the device and change its token to new canonical id:
+                    installation = clientInstallationService.findInstallationForVariantByDeviceToken(variantID,registrationIDs.get(i));
+                    installation.setDeviceToken(canonicalRegId);
+
+                    //update installation with the new token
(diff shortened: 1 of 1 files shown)


matzew:
@secondsun can you take a look ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/GCMPushNotificationSender.java
```diff
@@ -159,19 +160,44 @@ private void cleanupInvalidRegistrationIDsForVariant(String variantID, Multicast
...
+            //but you will get a new registration id with the name canonical id. Which mean, the registration id you sent the message to has 
+            //been changed to this canonical id, so change it on your server side as well.
+            
+            //check if current index of result has canonical id
+            String canonicalRegId = result.getCanonicalRegistrationId();
+            if (canonicalRegId != null) {
+                // same device has more than one registration id: update it
+              
+                // find the installation and change its token to new canonical id
+                Installation installation = clientInstallationService.findInstallationForVariantByDeviceToken(variantID,registrationIDs.get(i));
```
matzew: I think you need to add an import for the `Installation` model

matzew: What happens if let's say two or three different registrationIDs result in the same canonicalID ?

Should we perform some sort of merge? as it's really just one device?

sinarz: good point.
canonicalIDs are registrationIDs (tokens). So if two or more different registrationIDs result in the same canonicalID means either google has made a mistake and create a registrationID for different devices or both registrationIDs are from same device and we need to merge them as u noted.

sinarz: I added import for `Installation `class.

matzew:
> canonicalIDs are registrationIDs (tokens). So if two or more different registrationIDs result in the same canonicalID means either google has made a mistake and create a registrationID for different devices or both registrationIDs are from same device and we need to merge them as u noted.

@secondsun any thoughts?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/GCMPushNotificationSender.java
secondsun: @matzew It means that the user is registered multiple times.  The canonical ID is the most recent registrationID issued to a device.

matzew: @secondsun sent a PR @sinarz's branch, where I add a check for that.

Is it really impossible? I wonder if it's possible that (due to errors) one device has multiple (outdated) registration ids

secondsun:
@matzew The code looks sane.  I double checked all of Google's docs and this looks like the correct way to do the correct thing.

matzew:
there is still a bug in the code


working on it now

matzew:
@sinarz can you merge my PR to your branch?

so we have all changes in this one

sinarz:
sure
I think your code has a bug in it since it checks whether the canonical id exists in our system only if canonical id is null. Ill fix it and add it in my branch.

matzew:
@sinarz looks like you did not merge my two commits, but just copied something similar to yoru branch

matzew:
updated PR is here: https://github.com/aerogear/aerogear-unifiedpush-server/pull/693

PR #702, opened 2016-03-26
Changing Push Message Type to TEXT instead of VARCHAR

sinarz:
The type VARCHAR that is used in database for sending push messages is not good because it does not support unicode characters and users cannot send messages in some languages to their clients. (For example: Arabic, Persian, Urdu, etc)
The options are using NVARCHAR, TEXT and LONGTEXT.
I changed the database column type to LONGTEXT since I am sending push messages to different users that are found from my own database from multiple queries with a complicated logic.
One of my push messages contained almost 40 million characters, including more than 20k user device tokens.

I also changed hibernate settings to let the database exceed the limit of 4500 characters.

After doing the changes, it both supports unicode characters and long texts.

### DIFF
--- a/model/jpa/testData/create_db_content.ddl
+++ b/model/jpa/testData/create_db_content.ddl
@@ -71,7 +71,7 @@ CREATE TABLE push_message_info (
   client_identifier  VARCHAR(255)  DEFAULT NULL,
   ip_address         VARCHAR(255)  DEFAULT NULL,
   push_application_id VARCHAR(255) NOT NULL,
-  raw_json_message    LONGTEXT DEFAULT NULL,
+  raw_json_message    CLOB DEFAULT NULL,
   submit_date        DATE          DEFAULT NULL,
   total_receivers    BIGINT       NOT NULL,
   app_open_counter    BIGINT      DEFAULT 0,
--- a/model/jpa/testData/create_db_content.ddl
+++ b/model/jpa/testData/create_db_content.ddl
@@ -71,7 +71,7 @@ CREATE TABLE push_message_info (
   client_identifier  VARCHAR(255)  DEFAULT NULL,
   ip_address         VARCHAR(255)  DEFAULT NULL,
   push_application_id VARCHAR(255) NOT NULL,
-  raw_json_message    VARCHAR(4500) DEFAULT NULL,
+  raw_json_message    LONGTEXT DEFAULT NULL,
   submit_date        DATE          DEFAULT NULL,
   total_receivers    BIGINT       NOT NULL,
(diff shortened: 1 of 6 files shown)


matzew:
Hi, @sinarz 

the Travis CI job failed:
```
Results :

Tests in error: 
   Â» DatabaseCreatorFailure DDL execution failed. DB URL: 'jdbc:derby:memory:tes...

```

sinarz:
@matzew, trying to figure out why it has an error and fix this...

sinarz:
@matzew 
I don't know what is wrong with this file:
2015-07-08-normalize-push-message-information.xml
I changed it to the version it was before i made any changes and it still gives Travis-CI Check Failed error.

matzew:
ok, thanks for looking at it.

After my vacation, I'll take a look here too!

On Wed, Mar 30, 2016 at 1:35 PM, sinarz <notifications@github.com> wrote:

> @matzew <https://github.com/matzew>
> I don't know what is wrong with this file:
> 2015-07-08-normalize-push-message-information.xml
> I changed it to the version it was before i made any changes and it still
> gives Travis-CI Check Failed error.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/aerogear/aerogear-unifiedpush-server/pull/702#issuecomment-203389809>
>



-- 
Matthias Wessendorf

model/jpa/testData/create_db_content.ddl
```diff
@@ -71,7 +71,7 @@ CREATE TABLE push_message_info (
   client_identifier  VARCHAR(255)  DEFAULT NULL,
   ip_address         VARCHAR(255)  DEFAULT NULL,
   push_application_id VARCHAR(255) NOT NULL,
-  raw_json_message    VARCHAR(4500) DEFAULT NULL,
+  raw_json_message    LONGTEXT DEFAULT NULL,
```
matzew: @sinarz I think this causing errors with Derby.. (that's used in here)

In the migration folder we actually do use embedded MySQL - perhaps worth to do the same here ?

matzew: See http://db.apache.org/derby/docs/10.9/ref/crefsqlj31068.html#crefsqlj31068 I dont' see LONGTEXT

sinarz:
@matzew thank you for the link you provided.
I checked the code and changed LONGTEXT to CLOB but Travis-CI build failed again.
I tried and changed it to VARCHAR(4500) and got a Travs CI build failed error too. Don't really get what is wrong with this. I'm trying to make it right in my forked repository and update it here when its working correctly.

matzew:
closing as suggested. @sinarz feel free to submit one, when you got it working

PR #706, opened 2016-04-15
using execption message for error callback

matzew:
* Providing details on GCM's own error message to the callback
* remove duplicate/useless logger from the catch block
* using sendNoRetry, since we aren't doing any retry, nor have we internally support for a retry


matzew:
@danielpassos mind taking a quick look ?

danielpassos:
:+1:

PR #707, opened 2016-04-20
TokenLoader improvements

matzew:
* updating serialID right up from, to it's value is correct in all blocks
* as well added some loggers

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java
@@ -134,6 +134,11 @@ public class TokenLoader {
                 String lastTokenInBatch = null;
                 int tokensLoaded = 0;
                 for (int batchNumber = 0; batchNumber < configuration.batchesToLoad(); batchNumber++) {
+
+                    // increasing the serial ID,
+                    // to make sure it's properly read from all block
+                    ++serialId;
+
                     Set<String> tokens = new TreeSet<String>();
                     for (int i = 0; i < configuration.batchSize() && tokenStream.next(); i++) {
                         lastTokenInBatch = tokenStream.get();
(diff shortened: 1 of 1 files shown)


matzew:
@lfryc @secondsun mind taking a 👀  ❓

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java
```diff
@@ -134,32 +134,37 @@ public void loadAndQueueTokenBatch(@Observes @Dequeue MessageHolderWithVariants
                 String lastTokenInBatch = null;
                 int tokensLoaded = 0;
                 for (int batchNumber = 0; batchNumber < configuration.batchesToLoad(); batchNumber++) {
+
+                    // increasing the serial ID,
+                    // to make sure it's properly read from all block
+                    ++serialId;
```
secondsun: Is it correct for serialId to be incremented if tokens.size() == 0?

matzew: yes, it's an increment, based on the batch.

Currently the increase only happens here:
* https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java#L144

With the consequence that in other blocks the value is different, like: 
* https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java#L147
* or here, were we added the new log -> https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/TokenLoader.java#L159

matzew: Would be nice to get @lfryc take a quick look here too, but I think the current way of increasing is wrong

matzew:
@secondsun it's also covered in tests, the `serialId`:
* https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/test/java/org/jboss/aerogear/unifiedpush/message/jms/TestTokenBatchDeduplication.java
* https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push/sender/src/test/java/org/jboss/aerogear/unifiedpush/message/jms/TestTokenLoaderTransactionFailForGCM.java#L133

merging ... ⏫

PR #712, opened 2016-05-04
Adding check if we should really try topic push

matzew:
Sending message to an alias, was not working


secondsun:
All the following work
# I should get a message and I get one
* Global push
* Topic Puch
* Alias Push
* Alias + topic push
# I should not get a message and I do not get one
* Push to unmatched Alias
* Push to Catgeory + unmatched Alias

So +1 LGTM

PR #713, opened 2016-05-09
GCM 3 topic support

matzew:
porting the GCM 3 topic work to master

@secondsun

PR #714, opened 2016-05-09
Fix batch issues

matzew:
When there are several rounds of dispatched tokens (e.g. split in more than 10 batches), the current implementation would sent one topic request, per iteration.

1-10 batches: one topic request 
11-20 batches: one topic request 
21-30 batches: one topic request 
etc

The logic here fixes it. It was introuduced in https://github.com/aerogear/aerogear-unifiedpush-server/pull/713, which does contain a WF-10 specific test

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
@@ -122,13 +122,26 @@ public class TokenLoader {
                 final Set<String> topics = new TreeSet<>();
                 final boolean isAndroid = variantType.equals(VariantType.ANDROID);
 
+                // the entire batch size
+                int batchesToLoad= configuration.batchesToLoad();
+
+                // Some checks for GCM, because of GCM-3 topics
                 if (isAndroid && TokenLoaderUtils.isGCMTopicRequest(criteria)) {
 
-                    // If we are able to do GCM topics, we
-                    // 1) find all topis
-                    topics.addAll(TokenLoaderUtils.extractGCMTopics(criteria, variant.getVariantID()));
+                    // If we are able to do push for GCM topics...
 
-                    // 2) as well as legacy tokens
+                    // 1)
+                    // find all topics, BUT only on the very first round of batches
+                    // otherwise after 10 (or what ever the max. is) another request would be sent to that topic
+                    if (serialId == 0) {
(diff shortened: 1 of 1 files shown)


matzew:
btw. in https://github.com/aerogear/aerogear-unifiedpush-server/pull/713 we also have a test, which actually tests this GCM TOPIC dispatching, across several batch rounds

@secondsun any thoughts ?

secondsun:
@matzew How could we verify this manually?  I'm fine with the test in 1.2.x for automation purposes but I would like to see the code breaking in 1.1.x before I merge.  (otherwise LGTM)

matzew:
Ok, it's generally the token loading, that's buggy w/o this fix.

Run the WF/UPS with these System properties

```
-Daerogear.android.batchSize=1 -Daerogear.android.batchesToLoad=1
```


These make sure there is only one batch (per iteration) and each batch is one token big.

Without the fix, the first batch (which are topics) are getting push, but the legacy tokens are not triggered, because the size of batches is out of sync.

Resulting in wrong push (no push for legacy tokens):
```
17:32:03,918 WARNING [org.jboss.aerogear.unifiedpush.message.token.TokenLoader] (Thread-30 (HornetQ-client-global-threads-1699159127)) Check your push query: Not a single token was loaded from the DB!
```


With the fix, we get push to the topic (first batch iteration) and to legacy tokens (see log below, two pushes a 1 regitrationID)

```
...
18:00:27,719 INFO  [org.jboss.aerogear.unifiedpush.message.sender.GCMPushNotificationSender] (Thread-109 (HornetQ-client-global-threads-1699159127)) Sent push notification to GCM topic: &#x2F;topics&#x2F;50fdef1f-e9f6-435a-911d-5cc111e71a5f
...
18:00:27,728 INFO  [org.jboss.aerogear.unifiedpush.message.sender.GCMPushNotificationSender] (Thread-149 (HornetQ-client-global-threads-1699159127)) Sent push notification to GCM Server for 1 registrationIDs
...
18:00:27,771 INFO  [org.jboss.aerogear.unifiedpush.message.sender.GCMPushNotificationSender] (Thread-148 (HornetQ-client-global-threads-1699159127)) Sent push notification to GCM Server for 1 registrationIDs
...
```
NOTE: I hve three devices registered, two of them with legacy tokens.

matzew:
merged to 1.1.x branch, here: https://github.com/aerogear/aerogear-unifiedpush-server/commit/85ad0fad9ea1a8320bb1f09731d902d815aedd63

PR #722, opened 2016-05-17
Adding one more test-case for InstanceID format

matzew:


matzew:
@secondsun any concerns on adding yet another test ?

PR #725, opened 2016-05-19
add a @NotNull to the variant name field - AGPUSH-1634

lholmquist:
This is for https://issues.jboss.org/browse/AGPUSH-1634

The admin UI has it;s own check to make sure the name field is filled in before submitting, so this bug would have never been hit using the UI.

@matzew 👀

### DIFF
--- a/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/Variant.java
+++ b/model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/Variant.java
@@ -16,6 +16,7 @@
  */
 package org.jboss.aerogear.unifiedpush.api;
 
+import javax.validation.constraints.NotNull;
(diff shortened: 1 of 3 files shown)


matzew:
I think tests, e.g. here:
https://github.com/aerogear/aerogear-unifiedpush-server/blob/cbe1eeadfb647e4f1b6c8335feabe0daede97dd3/model/jpa/src/test/java/org/jboss/aerogear/unifiedpush/jpa/InstallationDaoTest.java#L598-L600

are still under the previous assumption, that a name is not required 😁

lholmquist:
@matzew i was wondering where those tests were  :)

lholmquist:
@matzew looks like i got all the tests now

matzew:
@lholmquist thanks a lot, dude! 🎉

PR #728, opened 2016-06-01
Remove Device form Topic based on previous categories

matskiv:
This work is base on [AGPUSH-1632](https://issues.jboss.org/browse/AGPUSH-1632) ticket.

This is my first PR for UPS server and code might not be up to best standards so I am looking forward to your comments.
Ping @matzew @secondsun

### DIFF
--- a/service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
+++ b/service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
@@ -1,0 +1,84 @@
+/**
+ * JBoss, Home of Professional Open Source
+ * Copyright Red Hat, Inc., and individual contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.aerogear.unifiedpush.service.util;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLEncoder;
+
+import org.jboss.aerogear.unifiedpush.api.AndroidVariant;
+import org.jboss.aerogear.unifiedpush.api.Category;
+import org.jboss.aerogear.unifiedpush.api.Installation;
+import org.jboss.aerogear.unifiedpush.utils.AeroGearLogger;
+
+/**
+ *  Utility used for Android installations which need to be unsubscribed from topics
+ */
+public class GCMTopicManager {
+
+    private final AeroGearLogger logger = AeroGearLogger.getInstance(GCMTopicManager.class);
+
+    // Instance ID API URL
+    public static final String IIDURL = "https://iid.googleapis.com/iid/v1/";
+
+    private AndroidVariant variant;
+
+    public GCMTopicManager(AndroidVariant variant) {
+        this.variant = variant;
+    }
+
+    /**
+     * Unsubscribes device from single category(topic)
+     *
+     * @param installation Installation object containing correct variant property of AndroidVariant class
+     * @param categoryToUnsubscribe category(topic) that device should be unsubscribed from
+     */
+    public void unsubscribe(Installation installation, Category categoryToUnsubscribe) {
+
+        String url = "";
+        try {
+            url = IIDURL + installation.getDeviceToken() + "/rel/topics/" + URLEncoder.encode(categoryToUnsubscribe.getName(), "UTF-8");
+        } catch (UnsupportedEncodingException e1) {
+            //
+        }
+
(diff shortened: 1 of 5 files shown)


secondsun:
This works most of the time, however if the device is subscribe to topics which UPS doesn't know about then the device will still get older messages.

TO demonstrate : 
Use this Android app : https://github.com/secondsun/unified-push-helloworld/tree/unsubscribe_topics

Start up an unpatched UPS instance
Setup your push app and replace the push-config in the android project
Start the android project
After you register select all of the topics from the options menu one at a time (there are three, sorry about the white on white)

You will notice that UPS thinks you are only subscribed to one topic, but you will receive messages for all topics.

Now patch the server and rerun the app.
You will reregister with 0 topics and UPS will unsubscribe you from the one topic it knew about.  You will still be subscribed to tow topics however.

You can work around this by querying the topics from the Google servers and updating based on THAT list and not on the list from the previous installation.

Hope this helps.

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
```diff
@@ -0,0 +1,84 @@
...
+            logger.fine("Unregistering device from topic was unsuccessfull");
+        }
+    }
+
+    /**
+     * Sends DELETE HTTP request to provided URL. Request is authorized using Google API key.
+     *
+     * @param urlS target URL string
+     */
+    protected int delete(String urlS) throws IOException {
```
matzew: should we do some better handling of IO issues?

secondsun: Our options are 1) schedule a retry or 2) make something pop up in the UI or 3) ??? (open for suggestions)

matskiv: > 1) schedule a retry
+1 (but I don't know how to implement that :( )

matskiv:
Thank you for review this guys. 
Nice catch @secondsun, I'll fix it as you suggested - by querying the topics from Google server and then unsubscribing.

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
matzew: > 1) schedule a retry or 2) make something pop up in the UI or 3) ??? (open for suggestions)

I think, we should not retry, just impriove the server side logging, in failure cases

Doing a pop up on the UI ... also not sure about it, because the update of the installations (re-registering) is triggering this, nothing really directly from the console.

I see the need for a better visiulation of general errors and better access to logs in future versions.

Do we have a ticket to improve our (real time) logs story ?

matskiv:
Now it queries Instance ID service for subscribed topics and after that unregisteres from those that are no longer relevant. 
Do you have suggestions on Exception handling here (instead of just logger.fine(..)) ?

service/pom.xml
```diff
@@ -49,6 +49,12 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
```
matzew: Do we need _really_ this ?  😄

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
```diff
@@ -0,0 +1,138 @@
...
+     * Unsubscribes device from single category(topic)
+     *
+     * @param installation Installation object containing correct variant property of AndroidVariant class
+     * @param categoryToUnsubscribe category(topic) that device should be unsubscribed from
+     */
+    public void unsubscribe(Installation installation, String categoryToUnsubscribe) {
+
+        String url = "";
+        try {
+            url = IIDURL + installation.getDeviceToken() + "/rel/topics/" + URLEncoder.encode(categoryToUnsubscribe, "UTF-8");
```
matzew: let's use `StandardCharsets.UTF_8` ?

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
```diff
@@ -0,0 +1,138 @@
...
+        } catch (UnsupportedEncodingException e1) {
+            //
+        }
+
+        try {
+            delete(url);
+        } catch (IOException e) {
+            logger.fine("Unregistering device from topic was unsuccessfull");
```
matzew: tjats ok...

service/pom.xml
matskiv: We already depend on it through gcm-server, so why not ? 
And if not, what should be used instead ? I tried jackson, but it wasn't... simple :smile:

matzew:
Can we have some unit tests?

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
matskiv: Ok, I'll fix it.

service/pom.xml
matzew: ok - fair point!

service/src/main/java/org/jboss/aerogear/unifiedpush/service/util/GCMTopicManager.java
matzew: cool

matskiv:
I'll add unit tests as soon as possible.

PR #729, opened 2016-06-03
Mongodb

pecsi:
use it if u want, but using hibernate ogm was imho not good idea


gunnarmorling:
Hey @pecsi, one of the Hibernate OGM devs here :)

I don't know anything about the context of this pull request, but I'm interested in your feedback. What was it you were trying to achieve here and why do you think using OGM was not a good idea eventually? Happy to learn from that. Thanks!

pecsi:
hi @gunnarmorling, i was not able to write aggregation and pagination queries in ogm so i used the underlining java mongo driver and when i replaced the relational hibernate with ogm it caused some issues with validation. Also the parser of native queries is stronger than mongo parser, so its not possible to prepare query in Robomongo or similar tool and then just paste it into the code.

matzew:
@pecsi I am not sure we will be able to merge thta, and replace MySQL :-(

gunnarmorling:
Thanks for your reply, @pecsi!

Could you elaborate a bit on the following:

> it caused some issues with validation

Do you mean Bean Validation here (JSR 303/349)? Or some other means of "validation"? What were these issues exactly?

> Also the parser of native queries is stronger than mongo parser,

Ah, that's great info. For sure it shouldn't be like that, can you give me examples of legal queries that the parser incorrectly refused?

Your feedback will be appreciated very much, so we can improve these things in Hibernate OGM. If you like, open an issue in our tracker (https://hibernate.atlassian.net/browse/OGM), and we can take a look quickly. Thanks again for your input, that's very helpful!

matzew:
closing as it's unlikely that it will be used... for now.

Thanks again @pecsi for contributing.  
In case we will move to Mongo, and away from MYSQL, this might become relevant again !

PR #731, opened 2016-06-21
Replacing AeroGearLogger with slf4j.

danbev:
Motivation:
To allow for more flexibility when it comes to the logging
implementation used in UPS we want to introduce slf4j. Using this
abstraction the choice of logging framework can be made at deployment
time.

Modifications:
Removed AeroGearLogger and introduced logging using the slf4j API.

Result:
All test pass and server logs are still produced as before.

JIRA:
https://issues.jboss.org/browse/AGPUSH-1086


danbev:
@matzew Would you mind taking a look at this? 
If it looks alright I'll create a PR against the 1.1.x-dev branch as well.

PR #741, opened 2016-07-04
Adding same validation as we have in standalone.xml

matzew:
Upating the health check to be identical to what we have for Openshift:
https://github.com/aerogear/openshift-origin-cartridge-aerogear-push/blob/master/versions/1.1.3/standalone/configuration/standalone.xml#L156


sebastienblanc:
:+1:

matzew:
and on master: https://github.com/aerogear/aerogear-unifiedpush-server/commit/0da5f12ed08c7bb18438069cf4ae2ee9a5f39847

PR #745, opened 2016-07-25
Add Webpush Variant to support Push API in Mozilla Firefox and Google Chrome

idelpivnitskiy:
After discussion in mailing list we decided to try create a Webpush variant for all browsers, which support Push API.

Current version supports sending push notifications just to Firefox.

To add Google Chrome we have to store auth tokens and determine the browser type in `WebpushPushNotificationSender.java`.

### DIFF
--- a/admin-ui/app/components/app-detail/include/analytics.js
+++ b/admin-ui/app/components/app-detail/include/analytics.js
@@ -72,6 +72,7 @@ angular.module('upsConsole')
         case 'windows_mpns': return 'Windows';
         case 'windows_wns': return 'Windows';
         case 'android': return 'Android';
+        case 'webPush': return 'WebPush';
         default: return variant.type;
       }
     }
--- a/admin-ui/app/components/app-detail/include/variantDetail.html
+++ b/admin-ui/app/components/app-detail/include/variantDetail.html
@@ -20,6 +20,9 @@
 <p ng-if="variant.type == 'windows_mpns'">Microsoft's Push Notification Service (MPNS) in Windows Phone will be used.
   <i>This type of Push Network is deprecated and will be removed in future versions</i>.
 </p>
+<p ng-if="variant.type == 'webPush'">Mozilla Push Service (MPS) for Firefox will be used.
+  <i>This is experimental support of Push API for web browsers</i>.
+</p>
 
 
 <!--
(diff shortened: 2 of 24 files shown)


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/WebpushPushNotificationSender.java
```diff
@@ -0,0 +1,67 @@
...
+
+@SenderType(VariantType.WEBPUSH)
+public class WebpushPushNotificationSender implements PushNotificationSender {
+    
+    private final Logger logger = LoggerFactory.getLogger(WebpushPushNotificationSender.class);
+    
+    private static final String MPS_URL = "https://updates.push.services.mozilla.com/wpush/v1/";
+
+    @Override
+    public void sendPushMessage(Variant variant, Collection<String> clientIdentifiers, UnifiedPushMessage pushMessage,
```
idelpivnitskiy: We can not determine the browser just from clientIdentifier, consequently we can not decide to which push notification provider we must send a push message (FCM or MPS).

We can use a hack and store the whole URL as a device token, like:
* https://gcm-http.googleapis.com/gcm/send/eIH9WwwyLzE:APA91bGTG1Nun97GHDs4afVq3kV0DWsERqf1_vm7hVf2EGnIbvv3UInCuCcqy4RIL7ZxyVOxE3g3_-V6G3RruRfvoNzNWql_YmrBywRUREwPoVaNxGJ-5sO1LhhnRKMzWPUIysdYg0Rh
* https://updates.push.services.mozilla.com/wpush/v1/gAAAAABXlQW2uvaJJk3Q6hey1cj2PjZYtGaDcY82DffVUF1OiV4Eu6SA1lds8jzKgZCR9JjIbioyv5jKwZQo2n6UxT8yRU3Es1qM2Fxmdv-p0cqGBhh4CjT5QNzlBAFRJ0OTLvisXB8e

Yes, we can determine appropriate push notification provider, but we also have to know "auth secret" and "public key" for each Installation to encrypt a payload of push message. So, I think it would be better to refactor PushNotificationSender to acquire `Collection<Installation> clients` instead of `Collection<String> clientIdentifiers`.

WDYT?

idelpivnitskiy:
@matzew I've also created a branch with name `webpush-variant` in `aerogear/aerogear-unifiedpush-server`: https://github.com/aerogear/aerogear-unifiedpush-server/tree/webpush-variant

Should this PR be closed?

matzew:
@idelpivnitskiy sorry for late reply - makes sense to move this to `webpush-variant` branch ? 

Or is this really isolated ?

idelpivnitskiy:
@matzew this code is already in `webpush-variant` branch. So, for me, this PR may be closed. And we may reopen PR from `webpush-variant` against `master`, just for code review.

matzew:
Ok, thanks for the update - will close this guy

PR #749, opened 2016-08-22
Refactoring and code improvements

idelpivnitskiy:
I've prepared a set of refactoring commits. Tried to make them as simple as possible. So it will be possible to remove some of unnecessary changes or cherry pick distinct commits.


matzew:
awesome!

PR #753, opened 2016-09-14
Added standard logger to RequestTransformer

nialldonnellyfh:
# Motivation

The RequestTransformer class is using a custom logger. To be more consistent, it should use the same logger as the other modules.


jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/RequestTransformer.java
```diff
@@ -23,16 +23,14 @@
 import com.github.fge.jsonpatch.JsonPatch;
 import com.github.fge.jsonpatch.JsonPatchException;
 import org.jboss.aerogear.unifiedpush.rest.util.transform.DynamicTransformer;
-
-import java.io.IOException;
```
matzew: do not remove the `IOException`

PR #756, opened 2016-09-19
comment out the section that triggers the 30 seconds polling

matzew:
For [AGPUSH-1895](https://issues.jboss.org/browse/AGPUSH-1895) we disable the polling

This also causes problems on screens that embed the AdminUI, since the interval is never cleared.

Also, polling is a bad technique, since the underlying SQL queries get executed over and over again, even when there is no change


matzew:
@lfryc If you don't mind, can you take a look ?

PR #757, opened 2016-09-19
Adding index on PushApplication#developer

gunnarmorling:
@matzew Hi, this should do the trick to add the index we discussed. Just edited this on GitHub, aven't tried it out locally ;)

### DIFF
--- a/model/jpa/src/main/resources/META-INF/orm.xml
+++ b/model/jpa/src/main/resources/META-INF/orm.xml
@@ -77,7 +77,9 @@ http://java.sun.com/xml/ns/persistence/orm_2_0.xsd"
         </attributes>
     </entity>
     <entity class="PushApplication" access="FIELD">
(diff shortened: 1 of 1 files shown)


matzew:
Hi @gunnarmorling! Thanks for that, but I am getting in TravisCI:
```
Error while reading JPA XML file: META-INF/orm.xml
```

we are on 2.0 😢 

https://github.com/gunnarmorling/aerogear-unifiedpush-server/blob/f2fdf0e9bc47a4a1da3dcf298ebfdec9d3c0fd1d/model/jpa/src/main/resources/META-INF/orm.xml#L23

gunnarmorling:
Doh, then indeed it seems you got to use hbm.xml or annotations :/

matzew:
for now, we close

thanks @gunnarmorling  - perhaps I will bug you in the future :)

PR #758, opened 2016-09-20
AGPUSH-1894 Using join instead of sub-select

gunnarmorling:
AGPUSH-1894

@matzew Can you give this one a try ?


gunnarmorling:
One interesting thing I came across is that there is no proper association from `PushMessageInformation` to `PushApplication` in the model, instead it references a natural key ("api key") instead.

I guess there are reasons for that design, though it may be a good idea to reconsider it and work with a regular `@ManyToOne` association (referencing the PK of `PushApplication`) instead: You can navigate the model (with the referenced application being loaded lazily if needed), use the proper JOIN syntax in JPQL etc.

Btw. I tried to model an object association using the current natural key via `property-ref` (see [here](https://docs.jboss.org/hibernate/orm/3.3/reference/en/html/mapping.html#mapping-declaration-manytoone)), but this is discouraged and seems to not work with lazy loading the reference, so I wouldn't go for that path.

matzew:
@gunnarmorling thanks! 💋  

Will take a 👀

matzew:
Thans, merge this to master and 1.x branch! 🎉

PR #770, opened 2016-12-23
Adding index for alias

matzew:
Adding index for alias.

*NOTE:* In theory this is a very small data set, so we likely don't gain too much from that...


matzew:
@odra mind reviewing ?

PR #771, opened 2016-12-23
Remove Variant criteria from topic check

matzew:
The `TokenLoader` class iterates over affected Vairants, per push job, and on EACH variant it checks if the request is a FCM/GCM topic request (e.g. does it have categories, or are there no criterias).

So, inside of a variant (used as topic any ways), it's not making sense, but instead returns false result.


matzew:
I need to write a test case...

matzew:
Testcase added

@odra @ziccardi mind to review ?

ziccardi:
LGTM

PR #773, opened 2017-01-20
Adding some more info what tool did send the request

matzew:
In our push metrics we collect the sender (tracked via `aerogear-sender` header or `user-agent`), but currently we are not showing the info (we used to do that in the old UI).

this PR displays the sender info on the activty log (tooltip in the IP-Address field)

It also adds it to the NotificationRouter log


matzew:
@odra @ziccardi mind doing a little review ?

ziccardi:
LGTM

PR #775, opened 2017-01-23
If there are no metrics available, there is no 0 index...

matzew:
If there is a variant, that has ZERO push notifications sent, clicking on the `Activity log` gives a JS error: `can not resolve property servedVariants of undefined`

This is due to the fact that, when there are no metric entires in the table, there is also nothing in index `0` ...

### DIFF
--- a/admin-ui/app/components/app-detail/include/activity.js
+++ b/admin-ui/app/components/app-detail/include/activity.js
@@ -75,7 +75,7 @@ angular.module('upsConsole')
         .then(function() {
           $log.debug('refreshed');
           // is the _last_ push job pending?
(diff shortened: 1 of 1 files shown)


matzew:
@josemigallas mind doing a quick review ?

josemigallas:
Sure @matzew

josemigallas:
@matzew Is this PR duplicated, by the way? https://github.com/aerogear/aerogear-unifiedpush-server/pull/776

matzew:
@josemigallas addressed

josemigallas:
Awesome

PR #777, opened 2017-01-24
Removing Openshift2 bits from the master branch

matzew:
The 1.1.x bits are still there, for OS2 - but master not!


sebastienblanc:
:+1:

PR #780, opened 2017-02-01
💄 Fix some user-visible spelling/grammar issues

grdryn:


grdryn:
@matzew Please take a look when you have a moment. Thanks! 🙏  

Should I PR against any other branches also, or do you cherry-pick at merge time?

admin-ui/app/scripts/directives.js
```diff
@@ -208,17 +208,17 @@ angular.module('upsConsole')
         show: true,
         text: {
           before: '<p>First you need to download add <code>unifiedpush-node-sender</code> as a <a ups-doc="sender-downloads-nodejs">dependency to your project</a>.</p>' +
-                  '<p>Then let\'s use following snippet in your Node.js code to enable push notification sending.</p>',
+                  '<p>Then let\'s use the following snippet in your Node.js code to enable push notification sending.</p>',
```
grdryn: The voice here is a little different than in other places: this is more conversational, where other places are more imperative (e.g. L220). Should I change this to "Use the following [...]" rather than "Then let's use the following [...]"?

matzew: @grdryn sure , it that's more aligned - sure !

matzew:
@grdryn thx for the PR

admin-ui/app/scripts/directives.js
grdryn: Ok cool, I'll update it to make that change then also. :)

grdryn: I've updated the commit to include this.

PR #782, opened 2017-02-01
Select fu

matzew:
replacing a sub-select w/ a join

@gunnarmorling mind checking  ? :) 

PS: of course, the test-case was tested with the old version as well :)


gunnarmorling:
Looks good to me 👍

PR #784, opened 2017-02-02
Duplicated requests in app details (tabs)

odra:
WIP

Trying to remove the duplicated requests when selecting a tab (variants, analytics, etc), it seems that both the click and the controller activation execute the same request(s) twice.

I removed the "ng-link" directive and added a check on the `$watch` function to verify if the new value is different from the old one

@lholmquist @sebastienblanc @edewit @matzew

### DIFF
--- a/admin-ui/app/components/app-detail/app-detail.html
+++ b/admin-ui/app/components/app-detail/app-detail.html
@@ -13,19 +13,18 @@
         </div>
         <h1>{{ appDetail.app.name }}</h1>
       </div><!-- app header -->
-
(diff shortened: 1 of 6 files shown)


matzew:
For me, I still see the duplicated requests - I don't really know too much of Angular 1.x

Perhaps @lfryc has a thought ?

matzew:
@lholmquist @sebastienblanc @edewit anyone able to take a look ?

josemigallas:
Taking the reins of this :horse:

PR #785, opened 2017-02-03
Breaking up java API, that was doing two things in one

matzew:


matzew:
@odra some refactoring

PR #786, opened 2017-02-04
Breaking up java API, that was doing two things in one

matzew:


matzew:
@gunnarmorling before both queries where in one method, which is now `deprecated` can calls into the two new methods.

I have a feeling that the `findMessageMetricsForPushApplicationByParams` is candidate for being slow, I wonder:
* any room for improvements? 
* would you also replace the [one service call](https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/metrics/PushMetricsEndpoint.java#L81-L82) with two individual ones? Mainly motivation would be performance,

model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPAPushMessageInformationDao.java
```diff
@@ -41,8 +41,27 @@
...
+        String baseQuery = "from PushMessageInformation pmi where pmi.pushApplicationId = :pushApplicationId";
+        if (search != null) {
+            baseQuery += " AND pmi.rawJsonMessage LIKE :search";
+        }
+        final String queryJPQL = "select pmi " + baseQuery + " ORDER BY pmi.submitDate " + ascendingOrDescending(ascending);
+
+        TypedQuery<PushMessageInformation> typedQuery = createQuery(queryJPQL)
+                .setParameter("pushApplicationId", pushApplicationId);
+        if (search != null) {
+            typedQuery.setParameter("search", "%" + search + "%");
```
gunnarmorling: This one may pose an issue, as the wildcard operator at the beginning of the search term likely will trigger a full-table scan. Can you resort to (vendor-specific) full-text search functions? Or you use Hibernate Search and Lucene :)

But it's all a bit guess work on my side, would be interesting to see some profiling results of where your execution actually is spent and how execution plans of the queries look like.

model/jpa/src/main/java/org/jboss/aerogear/unifiedpush/jpa/dao/impl/JPAPushMessageInformationDao.java
```diff
@@ -57,31 +76,27 @@ public long getNumberOfPushMessagesForVariant(String variantID) {
...
-        if (search != null) {
-            typedQuery.setParameter("search", "%" + search + "%");
-        }
-        typedQuery.setFirstResult(page * pageSize).setMaxResults(pageSize);
-        List<PushMessageInformation> pushMessageInformationList = typedQuery.getResultList();
-
-        Query metricsQuery = createUntypedQuery(metricsJPQL).setParameter("pushApplicationId", pushApplicationId);
+        final Query metricsQuery = createUntypedQuery(metricsJPQL).setParameter("pushApplicationId", pushApplicationId);
         if (search != null) {
             metricsQuery.setParameter("search", "%" + search + "%");
```
gunnarmorling: Same here as above likely.

matzew:
thanks - the like will be fixed

PR #788, opened 2017-02-11
Variant dispatch fix

matzew:
Some improvements on push msg dispatching
* no DB lookup (per ID), via service; Instead check child's if ID is present
* use Set, instead of List, for single occurrence
* fix broken test, which did match broken behavior :dagger:


matzew:
merged

PR #794, opened 2017-02-18
Batch endpoint

matzew:
fast track gazillions of push :email: for be delivered in one batch :mailbox: for the same Push-Application

### DIFF
--- a/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
+++ b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
@@ -30,23 +30,28 @@ import javax.ws.rs.core.Response.Status;
 import com.qmino.miredot.annotations.BodyType;
 import com.qmino.miredot.annotations.ReturnType;
 import org.jboss.aerogear.unifiedpush.api.PushApplication;
+import org.jboss.aerogear.unifiedpush.message.PushMessageBatchDispatcher;
 import org.jboss.aerogear.unifiedpush.rest.EmptyJSON;
 import org.jboss.aerogear.unifiedpush.message.InternalUnifiedPushMessage;
-import org.jboss.aerogear.unifiedpush.message.NotificationRouter;
 import org.jboss.aerogear.unifiedpush.rest.util.HttpBasicHelper;
 import org.jboss.aerogear.unifiedpush.rest.util.HttpRequestUtil;
 import org.jboss.aerogear.unifiedpush.service.PushApplicationService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
+import java.util.List;
+
 @Path("/sender")
 public class PushNotificationSenderEndpoint {
 
     private final Logger logger = LoggerFactory.getLogger(PushNotificationSenderEndpoint.class);
+
+    @Inject
+    private PushMessageBatchDispatcher dispatcher;
+
     @Inject
     private PushApplicationService pushApplicationService;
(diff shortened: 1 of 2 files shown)


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/PushMessageBatchDispatcher.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Helper method, fired by the endpoint, to allow async processing of large push message collections
+     * @param pushApplication push application receiving the message
+     * @param messages collection of messages for the given push application
+     */
+    @Asynchronous
+    public void dispatchPushMessageBatch(final PushApplication pushApplication, final List<InternalUnifiedPushMessage> messages) {
+
+        for (InternalUnifiedPushMessage msg : messages) {
```
matzew: perhaps we could pause after a configurable number of batches, like:

```   for() {
      ... current proposed code
        if (i % 5000 == 0) {
          Thread.sleep(1000); // wait a second
   }
}
```

matzew: but than, it's all going to JMS, which is already configured to run in batches

camilamacedo86: I'd like to suggest to you use Java8 here. 

messages.stream().forEach(msg->msg.submit(pushApplication, msg);

Also, I don't know how is the setup of the VM parameters and the production environment. However, I think that is a good idea try to test check the behaviour here with the use of parallel stream API to improve the performance.  

messages.parallelStream().forEach(msg->msg.submit(pushApplication, msg);

Also, has two for to iterate the same list. Is not possible do just one iteration to win performance?

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
```diff
@@ -128,4 +180,21 @@ private PushApplication loadPushApplicationWhenAuthorized(HttpServletRequest req
...
+    private void processPushMessages(final HttpServletRequest request, final PushApplication pushApplication, final List<InternalUnifiedPushMessage> messages) {
+
+        // read some metadata about the submitting client
+        final String ipAddress = HttpRequestUtil.extractIPAddress(request);
+        final String clientSender = HttpRequestUtil.extractAeroGearSenderInformation(request);
+
+        // apply client metada
+        for (InternalUnifiedPushMessage msg : messages) {
+            msg.setIpAddress(ipAddress);
+            msg.setClientIdentifier(clientSender);
```
camilamacedo86: Here, why don't use Java8 to do this interaction and send here the message directly for don't be required do the same for again?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/PushMessageBatchDispatcher.java
matzew: Hrm, yeah, on master we do use java8, on the 1.1.x breanch not, 

regarding parallel streams: no-go, since that's Fork-Join, which is not supported in JavaEE6, nor in JavaEE7.

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/sender/PushNotificationSenderEndpoint.java
matzew: as set, we can't do java8 on that this branch,

PR #795, opened 2017-02-20
Renaming desc. of the APNs mock

matzew:


kpiwko:
Thanks for this improvement!

PR #800, opened 2017-02-23
Add socks proxy support to APNS and health endpoint

PhilipGough:
ping @matzew can you review?

### DIFF
--- a/service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/health/Ping.java
+++ b/service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/health/Ping.java
@@ -16,6 +16,8 @@
  */
 package org.jboss.aerogear.unifiedpush.service.impl.health;
 
+import org.jboss.aerogear.unifiedpush.service.proxy.ProxyConfiguration;
+
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/jboss/aerogear/unifiedpush/service/proxy/ProxyConfiguration.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.annotation.PostConstruct;
+import javax.ejb.Singleton;
+import javax.ejb.Startup;
+import java.net.Authenticator;
+import java.net.InetSocketAddress;
+import java.net.PasswordAuthentication;
+import java.net.Proxy;
+
+@Startup
+@Singleton
```
matzew: move the annotations above the `public class...` line

matzew:
besides the one comment looks good.

One more thing, can you add the new ENVs to this document:
https://github.com/aerogear/aerogear.org/blob/master/docs/unifiedpush/ups_userguide/server-installation_properties.asciidoc#system-properties-for-the-unifiedpush-server

NOTE: there is only System properties, so feel free to create a new `Environment Variables` section _above_

service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/health/Ping.java
```diff
@@ -26,7 +28,12 @@ public static boolean isReachable(String host, int port) {
...
         boolean reachable;
         try {
-            socket = new Socket();
+
+            if (ProxyConfiguration.hasSocks()) {
+                socket = new Socket(ProxyConfiguration.socks());
+            } else {
+                socket = new Socket();
+            }
             socket.connect(new InetSocketAddress(host, port), 2000);
```
matzew: @PhilipGough is the 2 second timeout for you OK as well ?

PhilipGough: Works great, returning in less than 200 typically

matzew: I saw the others, that worked ;-) return in under 100 ms

PR #801, opened 2017-02-23
Fcm fix and timeout

matzew:
as well IP address validation needed pattern (OpenJDK)


matzew:
forward port from 1.1.x branch

PR #804, opened 2017-03-21
Only one request when changing between tabs

josemigallas:
This PR solves the ticket: [AGPUSH-2052](https://issues.jboss.org/browse/AGPUSH-2052)

It is an alternative for #784.

#### Reason
After a lot of debugging, the problem _seems to be_ the "ngLink" directive, that generates an url starting with `.` instead of `#`. 

#### Solution
Although one solution would be changing all `.` for `#` in the declaration of the directive (`ngLinkDirective` function in `app/scripts/angular.router.es5.js`) I find more readable and clean to simply specify the reference.

I think the best way to deal with this behaviour would be to create a method `setTab` to the appDetail controller. By updating the `tab` property this way, the component would be shown and refreshed. *However*, the current implementation requires the URL to be the one to update in order to fetch the components data.

### DIFF
--- a/admin-ui/app/components/app-detail/app-detail.html
+++ b/admin-ui/app/components/app-detail/app-detail.html
@@ -16,16 +16,16 @@
 
       <ul class="nav nav-tabs">
         <li ng-class="{ active: appDetail.tab == 'analytics' }" ng-if="appDetail.app.variants.length">
-          <a id="analytics-tab" ng-link="appDetail({app: appDetail.app.pushApplicationID, tab: 'analytics'})">Analytics</a>
+          <a id="analytics-tab" href="#/app/{{appDetail.app.pushApplicationID}}/analytics">Analytics</a>
         </li>
         <li ng-class="{ active: appDetail.tab == 'variants' }">
-          <a id="variants-tab" ng-link="appDetail({app: appDetail.app.pushApplicationID, tab: 'variants'})">Variants</a>
+          <a id="variants-tab" href="#/app/{{appDetail.app.pushApplicationID}}/variants">Variants</a>
         </li>
         <li ng-class="{ active: appDetail.tab == 'sender' }" ng-if="appDetail.app.variants.length">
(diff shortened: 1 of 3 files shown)


josemigallas:
#784

matzew:
I am not able to click on the `Activity log` tab

matzew:
@edewit @lholmquist @odra can one of you guys take a look as well?
Besides actual testing, in terms of code-change, I am not really familiar w/ Angular at all

josemigallas:
@matzew It's solved now. I mistyped the href and point both sender and activity tabs to sender.

lholmquist:
I haven't tested yet,  just wanted to comment on the changes.  It does seem a little hacky since it goes a little against the way i think angular wants you to code,  but at the same time,  the router that is being used here is no longer supported.  Its also a file that has had updates made to it,  and not part of any dependency manager(like bower or npm),  so i think if these changes work, then they should be ok.  Unless someone wants to re-write the whole UI over again :)

that might be a better option if the decision is made to move this(the admin UI) to its own repo,  but thats a whole other issue

matzew:
I noticed that with this change clicking on the current TAB has no effect anymore.

I am not really sure, but I guess it was designed to fetch data when clicking on the (current) tab. 

I see the dobule requests are gone, when transitioning from A->B, but no request is triggered on clicking on the current TAB.

@edewit any comments?

josemigallas:
@lholmquist I think the whole UI is a bit tricky, that's why a think it should be rewritten too.

matzew:
Not sure why the build fails... hrm

josemigallas:
Me neither, it's building locally.

matzew:
Hrm,

the admin ui module take oddly long, in the maven build:
```
[INFO] UnifiedPush Admin UI ............................... SUCCESS [08:31 min]
```

rebuilding to double check

matzew:
second build of the branch went well :-)

matzew:
@josemigallas do you mind creating a new, different PR for 1.1.x-dev brnach, of this fix ? 

thanks!

matzew:
build fails due to travis time out, known issue

PR #805, opened 2017-03-21
Improves readme info about environment setup

josemigallas:
This PR solves the ticket: https://issues.jboss.org/browse/AGPUSH-2066

### DIFF
--- a/admin-ui/README.md
+++ b/admin-ui/README.md
@@ -39,7 +39,7 @@ ag-push
 `ag-push.war` can be deployed directly but in order for Grunt to work properly, `ag-push` must be deployed instead. To do that, simply rename it to `ag-push.war` and move it to the `standalone/deployments` directory:
 ```
 cd $SERVER_HOME/standalone/deployments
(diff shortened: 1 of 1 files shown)


admin-ui/README.md
```diff
@@ -2,38 +2,55 @@
...
+```
+That's because you did not deploy the correct file. After building the whole project with `mvn`, you will have this two files located in `aerogear-unifiedpush-server/servers/ups-wildfly/target/`:
+```
+ag-push.war
+ag-push
+```
+`ag-push.war` can be deployed directly but in order for Grunt to work properly, `ag-push` must be deployed instead. To do that, simply rename it to `ag-push.war` and move it to the `standalone/deployments` directory:
+```
+cd $SERVER_HOME/standalone/deployments
+cp path/to/aerogear-unifiedpush-server/servers/ups-wildfly/target/ag-push ag-push.war
```
lholmquist: I think that needs to be "cp -r" since you are copying a directory

josemigallas: You are completely right :+1:

lholmquist:
@josemigallas i think this is looking pretty good. 👍

matzew:
:tada:

PR #806, opened 2017-03-24
Only one request when changing between tabs

josemigallas:
This PR solves the ticket: [AGPUSH-2052](https://issues.jboss.org/browse/AGPUSH-2052)

It is an alternative for #784.
It is the same as of #804 

#### Reason
After a lot of debugging, the problem _seems to be_ the "ngLink" directive, that generates an url starting with `.` instead of `#`. 

#### Solution
Although one solution would be changing all `.` for `#` in the declaration of the directive (`ngLinkDirective` function in `app/scripts/angular.router.es5.js`) I find more readable and clean to simply specify the reference.

I think the best way to deal with this behaviour would be to create a method `setTab` to the appDetail controller. By updating the `tab` property this way, the component would be shown and refreshed. *However*, the current implementation requires the URL to be the one to update in order to fetch the components data.


matzew:
@josemigallas there is something wrong w/ this PR - see the number of commits.

Was it originally planed for `master` branch ?

josemigallas:
It didn't look so bad in my local repo, fixing it now.

josemigallas:
So I did merge with master before sending the PR, oops! Closing this and sending new PR.

PR #807, opened 2017-03-28
:iphone: :loudspeaker: HTTP/2 work for APNs

matzew:
Java7 / EAP 6.4 version of simple HTTP/2 APNs work.

Auth is still done via TLS, no token-auth yet


matzew:
@ziccardi can you review this ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+
+    @Inject
+    private ClientInstallationService clientInstallationService;
+    private final ConcurrentSkipListSet<String> invalidTokens = new ConcurrentSkipListSet();
+
+
+    @Override
+    public void sendPushMessage(final Variant variant, final Collection<String> tokens, final UnifiedPushMessage pushMessage, final String pushMessageInformationId, final NotificationSenderCallback senderCallback) {
+        // no need to send empty list
+        if (tokens.isEmpty()) {
```
ziccardi: Can tokens be null?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+    private static final String customAerogearApnsPushHost = tryGetProperty(CUSTOM_AEROGEAR_APNS_PUSH_HOST);
+    private static final Integer customAerogearApnsPushPort = tryGetIntegerProperty(CUSTOM_AEROGEAR_APNS_PUSH_PORT);
+
+    @Inject
+    private ClientInstallationService clientInstallationService;
+    private final ConcurrentSkipListSet<String> invalidTokens = new ConcurrentSkipListSet();
+
+
+    @Override
+    public void sendPushMessage(final Variant variant, final Collection<String> tokens, final UnifiedPushMessage pushMessage, final String pushMessageInformationId, final NotificationSenderCallback senderCallback) {
```
ziccardi: Since this is a public message, I think we should add sanity checks on the parameters

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
matzew: no, it's always a collection, but not null

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
matzew: yeah, we can do that, but they are just called by one class, which has reasonable arguments for it

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        final iOSVariant iOSVariant = (iOSVariant) variant;
+
+        final String payload;
+        {
+            try {
+                payload = createPushPayload(pushMessage.getMessage(), pushMessageInformationId);
+            } catch (IllegalArgumentException iae) {
+                logger.info(iae.getMessage());
```
ziccardi: I think we should log the exception itself to have the stacktrace

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+
+        for (final String token : tokens) {
+            final AeroGearApnsPushNotification pushNotification = new AeroGearApnsPushNotification(token, payload);
+            final Future<PushNotificationResponse<AeroGearApnsPushNotification>> notificationSendFuture = apnsClient.sendNotification(pushNotification);
+            try {
+                handlePushNotificationResponsePerToken(notificationSendFuture.get());
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
ziccardi: Would be better to use the logger

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+            } catch (ExecutionException e) {
+                e.printStackTrace();
```
ziccardi: See previous comment

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+                return;
+            }
+        }
+
+        final ApnsClient apnsClient = buildApnsClient(iOSVariant);
+
+        // connect:
+        final Future<Void> connectFuture = connectToDestinations(iOSVariant, apnsClient);
+        try {
+            connectFuture.await();
```
ziccardi: Listener should be preferred to avoid blocking

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        // we have managed to send all messages ;-)
+        senderCallback.onSuccess();
+
+        // stop connection
+        final Future<Void> disconnectFuture = apnsClient.disconnect();
+        try {
+            disconnectFuture.await();
```
ziccardi: Listener should be generally preferred to 'await'

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+            clientInstallationService.removeInstallationsForVariantByDeviceTokens(iOSVariant.getVariantID(), invalidTokens);
+
+
+        } catch (InterruptedException e) {
+            logger.info("Error disconnecting from APNs", e);
+        } finally {
```
ziccardi: Nothing to be done in the finally block? If yes, this should be removed

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+        final String deviceToken = pushNotificationResponse.getPushNotification().getToken();
+
+        if (pushNotificationResponse.isAccepted()) {
+            logger.trace(String.format("Push notification for '%s' (payload=%s)", deviceToken, pushNotificationResponse.getPushNotification().getPayload()));
+        } else {
+            final String rejectReason = pushNotificationResponse.getRejectionReason();
+
+            // token is either invalid, or did just expire
+            if ((pushNotificationResponse.getTokenInvalidationTimestamp() != null) || ("BadDeviceToken".equals(rejectReason))) {
+                logger.warn(rejectReason + ", removing token: " + deviceToken);
```
ziccardi: Shouldn't this be an INFO?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+                        .setClientCredentials(stream, iOSVariant.getPassphrase())
+                        .build();
+
+                // release the stream
+                stream.close();
+            } catch (Exception e) {
+                logger.error("Error reading certificate", e);
+
+                // indicating an incomplete service
+                return null;
```
ziccardi: Wouldn't be better to have an exception here?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+            return apnsClient;
+        }
+        // null if, why ever, there was no cert/passphrase
+        return null;
```
ziccardi: See my previous comment

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
matzew: @ziccardi Good point. However, we need to wait until the connection is established, in order to send the notifications.

Now, what we could do, in the listener, we can check the status of the future, and start sending, when connection is established.

WDYT ?

ziccardi: Yeah, that's what I was thinking.

matzew: So, I'd need to move the "send" inside of that callback, because otherwise there is different exception (e.g. ClientNotConnected).

let me update

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+
+        // connect and move on with sending in the background:
+        final Future<Void> connectFuture = connectToDestinations(iOSVariant, apnsClient);
+        connectFuture.addListener(new GenericFutureListener<Future<? super Void>>() {
+            @Override
+            public void operationComplete(Future<? super Void> future) throws Exception {
+
+                if (!future.isSuccess()) {
```
ziccardi: I think this should be reverted to 
if (future.isSuccess()) {
...
} else

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -0,0 +1,245 @@
...
+
+                                if (future.isSuccess()) {
+                                    handlePushNotificationResponsePerToken(notificationSendFuture.get());
+                                }
+                            }
+                        });
+                    }
+
+                    // we have managed to send all messages ;-)
+                    senderCallback.onSuccess();
```
ziccardi: The sendNotification is asynchronous, so when we arrived here we still don't know if we successfully sent the message or not. Is this correct?

matzew: at this point we are sending, and we generally are OK if the connection is made, payload is correct. we don't do fine tracking like "3 tokens invalid" etc

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
ziccardi: Ok, if we are sure that we won't invoke this method from other classes in the future. Otherwise I think the sanity check should be added here.

ziccardi:
LGTM. Just some comment

PR #809, opened 2017-03-30
Token cleanup and apns connection cache

matzew:
* remove invalid tokens (was missing)
* cache per iOS variant, per node in cluster (APNs wants us to stay _busy_ connections open)
* inactive connections, are cleaned by, from the cache

### DIFF
--- a/push/sender/pom.xml
+++ b/push/sender/pom.xml
@@ -91,6 +91,12 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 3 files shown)


ziccardi:
LGTM

matzew:
@ziccardi did you see the `synchronized` method on the sender ? I made it private, non-static, since we cache connection per variant

PR #811, opened 2017-03-31
Synchronized connection creation

matzew:
avoid creating multiple connections, when actively needing one, per (cluster) node

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
@@ -26,12 +26,12 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.PreDestroy;
-import javax.enterprise.context.ApplicationScoped;
+import javax.ejb.Singleton;
 import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 
-@ApplicationScoped
+@Singleton
 public class SimpleApnsClientCache {
 
     private final Logger logger = LoggerFactory.getLogger(SimpleApnsClientCache.class);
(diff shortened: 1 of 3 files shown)


matzew:
@ziccardi any comments ?

matzew:
@ziccardi can we merge this one ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -44,10 +46,11 @@
 import static org.jboss.aerogear.unifiedpush.message.util.ConfigurationUtils.tryGetIntegerProperty;
 import static org.jboss.aerogear.unifiedpush.message.util.ConfigurationUtils.tryGetProperty;
 
+@Stateless
 @SenderType(VariantType.IOS)
 public class PushyApnsSender implements PushNotificationSender {
 
-    private final Logger logger = LoggerFactory.getLogger(PushyApnsSender.class);
+    private static final Logger logger = LoggerFactory.getLogger(PushyApnsSender.class);
```
ziccardi: Static reference to slf4j should be avoided: https://wiki.apache.org/commons/Logging/StaticLog

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
```diff
@@ -70,12 +70,29 @@ public void operationComplete(Future<? super Void> future) throws Exception {
...
+        }
     }
 
-    public void putApnsClientForVariantID(final String variantID, final ApnsClient apnsClient) {
-        apnsClientExpiringMap.put(variantID, apnsClient);
+    private void putApnsClientForVariantID(final String variantID, final ApnsClient apnsClient) {
+        final ApnsClient client = apnsClientExpiringMap.putIfAbsent(variantID, apnsClient);
+        if (client != null) {
+            logger.warn("duplicate connection in pool, shutting it down");
+            tearDownApnsHttp2Connection(client);
```
ziccardi: putIfAbsent returns the **previous** value associated with the key. Shouldn't we shutdown `apnsClient` instead of `client`?

ziccardi:
Added a couple of comments

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
matzew: @ziccardi the previous (`client`) is replace with `apnsClient`, right ? that means the cache will contain the latest, so we tear down the previously cached value.  Or does not not make sense ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
matzew: ah, done by accident - will update

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
ziccardi: In my understanding the new value is inserted only if the key is not already associated with another value.
So, the returned value is the old one and, in the map, the old one is present.

> If the specified key is not already associated with a value, associate it with the given value.

matzew:
@ziccardi addressed them - good for merging ?

ziccardi:
LGTM

PR #812, opened 2017-04-06
Remove appcelerator references

sedroche:
# Motivation
As per ticket [FH-3250](https://issues.jboss.org/browse/FH-3250) we are no longer supporting Titanium/appcelerator apps. Removing them from ups also.


matzew:
If the FeedHenry project no longer supports Titanium/appcelerator apps, that does not automatically mean the same is true for the AeroGear community. We do have users, and contributors! that are doing Titanium.

matzew:
:-1:  on this

PR #814, opened 2017-04-10
Using latest of pushy

matzew:
Using latest of pushy and the netty-tcnative-boringssl-static lib

The other versions are leveraging tomcat (fork) classes, so there is chance of clash.

Now, we do not need to do any exclude of the Jboss Web layer anymore


ziccardi:
LGTM. No version bump?

matzew:
> No version bump? 

NOOOOOOOOOOO :-) we do _proper_ release management ;-) and only bump on releases, via maven plugin :)

PR #815, opened 2017-04-11
only stash connected clients

matzew:
generally we try to send only, when client is connected:
https://github.com/aerogear/aerogear-unifiedpush-server/blob/1.1.x-dev/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java#L99-L119

but, if - for what ever reason, we can not connect, let's not stash a disconnected client object...

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
@@ -77,7 +77,10 @@ public class SimpleApnsClientCache {
             logger.debug(String.format("no cached connection for %s, establishing it", variantID));
             synchronized (apnsClientExpiringMap) {
                 client = constructor.construct();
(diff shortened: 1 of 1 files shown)


matzew:
@ziccardi can you review ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
```diff
@@ -77,7 +77,10 @@ public ApnsClient getApnsClientForVariantID(final String variantID, final Servic
             logger.debug(String.format("no cached connection for %s, establishing it", variantID));
```
ziccardi: We are using SLF4J. 
We could rewrite this as:
`logger.debug("no cached connection for {}, establishing it", variantID);`

That would be more performant since no interpolation is performed if the log is not printed

ziccardi:
LGTM. Just one comment

matzew:
Thx, @ziccardi I will send a new PR w/ some more logger stuff

PR #816, opened 2017-04-11
Socks proxy

matzew:
@maleck13 @PhilipGough 

We have changed the library to connect to APNs, Changes:
* Using HTTP/2 endpoints
* using pushy lib (based on Netty 4.1.x)

This PR is adding some proxy support, but I have a few questions:
* I am using a `Socks5ProxyHandlerFactory`, but the library also offers a `Socks4ProxyHandlerFactory` - do we need to handle the case? or is Socks V5 just fine?
* In the proxy config clazz, I do see there is no user/password for Socks - do we not support that ? I only see them for HTTP proxy..., correct ? 
* for HTTP proxy, we don't do anything fancy, as we try, on RT, to set the HTTP proxy globally, using EVN/SYS properties, right ?


PhilipGough:
Reviewed. Looks good

PhilipGough:
1. I think thats fine
1. N/A, this is just a TCP tunnel from my understanding
1. Correct

PR #818, opened 2017-04-13
Modify indentation

dimitraz:
Changing the compose file from the `aerogear/unifiedpush-wildfly-dev` to the `aerogear/unifiedpush-wildfly` image didn't work until I corrected the indentation. Updating here as well for consistency. 

@matzew can you review?


matzew:
@dimitraz cool, thanks !

I just landed it :tada:

PR #819, opened 2017-04-15
APNs multi staging fixes 

matzew:
This PR addresses two things:

* when network options of a iOS variant change from DEV->PROD (or vice versa), the cached connection is incorrect, since we only use the Variant_ID as the key. We need to a add a new identified to the `VARIANT_ID`, like `-prod` or `-dev`

* adding Default Topic for APNs (bundle ID)
When an application/bundle_ID (inside of Apple's UI) is configured for multiple stages (e.g. DEV/PROD), we do need to provide a (default) topic, which matches the Apple bundle ID. We read that from the given `.p12` file, if present. Otherwise, `null` is used - which is fine for _single_ topic certificats (e.g. Only one, DEV _or_ PROD is enabled), and on `DEVELOPMENT APNS` - unfortunately the `PRODUCTION` endpoint is a bit inconsistent...

Besides that, removing a dependency, which is automatically get from Pushy `0.9.3` and some simple LOGGER format changes.


matzew:
@ziccardi Can you review this ?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.ByteArrayInputStream;
+import java.security.KeyStore;
+import java.security.cert.X509Certificate;
+import java.util.Enumeration;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public final class ApnsUtil {
+
+    private static final String ALGORITHM = "PKCS12";
```
ziccardi: PKCS12 is is a KEYSTORE TYPE. I think it should be renamed

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -0,0 +1,69 @@
...
+    private static final String TOPIC_PATTERN = ".*UID=([^,]+).*";
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ApnsUtil.class);
```
ziccardi: Slf4j logger should not be static

matzew: yeah, but it's a util, only having static methods :thinking:

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private ApnsUtil() {
+    }
+
+    public static String readDefaultTopic(final byte[] cert, final String password) {
```
ziccardi: `cert` is an array of byte representing a `keystore`, not a certificate. I think naming this `cert` is misleading.
Passwords should never be stored into Strings (strings are pooled and are immutable, so they have big chance to remain in memory for a long time)

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
ziccardi: My bad. Ignore this than ;)

ziccardi:
Added a few comments

ziccardi:
LGTM

PR #820, opened 2017-04-20
Moving the success cb up

matzew:
Once we are sure we are connected, but _before_ the streaming of all tokens/payload pairs, we now invoke the `onSuccess` handler.

We are able to do because we are _successfully_ connected to APNs.

If one of the tokens in the iteration is a bogus one, we generally not flagging this as `failure`... since that also makes no sense (e.g. if one token out of 10.000 is bogus, the transaction is _NOT_ failed).

The `onError` handle is invoked, if we are not connected to APNs, which really indicates a failure, an error

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
@@ -102,6 +102,9 @@ public class PushyApnsSender implements PushNotificationSender {
 
         if (apnsClient.isConnected()) {
 
+            // we have managed to connect and will send tokens ;-)
+            senderCallback.onSuccess();
+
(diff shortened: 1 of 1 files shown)


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -192,8 +193,9 @@ public ApnsClient construct() {
                 // APNS client has auto-reconnect, but let's log when that happens
                 apnsClient.getReconnectionFuture().addListener(new GenericFutureListener<Future<? super Void>>() {
                     @Override
+
```
josemigallas: Don't add extra lines

matzew: @josemigallas fixed

PR #826, opened 2017-05-08
Fixes UPS GUI links broken in studio

josemigallas:
This PR solves the ticket: [AGPUSH-15180](https://issues.jboss.org/browse/RHMAP-15180)

#### Reason
In a previous PR (#810) the ngLink directive that was causing a double HTTP request was changed into a simple href property. This might have caused the current issue.

#### Solution
Revert the changes made in #810.

#### Collateral Damage
Reverting to #810 will re-open the double requests issue so this fix should be only temporal until a better solution is found.


matzew:
merge it !

matzew:
mind sending a PR with this change to the master as well ? 

since there we do need the undo as well :+1:

josemigallas:
Done @matzew #828

PR #827, opened 2017-05-08
HTTP Proxy support for Pushy

PhilipGough:
http proxy support for the pushy lib for ios apns. Update the health check ping functionality and continue support for socks proxy

ping @matzew

### DIFF
--- a/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
+++ b/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
@@ -19,6 +19,7 @@ package org.jboss.aerogear.unifiedpush.message.sender;
 import com.relayrides.pushy.apns.ApnsClient;
 import com.relayrides.pushy.apns.ApnsClientBuilder;
 import com.relayrides.pushy.apns.PushNotificationResponse;
+import com.relayrides.pushy.apns.proxy.HttpProxyHandlerFactory;
 import com.relayrides.pushy.apns.proxy.Socks5ProxyHandlerFactory;
 import com.relayrides.pushy.apns.util.ApnsPayloadBuilder;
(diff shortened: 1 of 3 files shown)


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/PushyApnsSender.java
```diff
@@ -214,9 +214,17 @@ private ApnsClient buildApnsClient(final iOSVariant iOSVariant) {
                 final ApnsClientBuilder builder = new ApnsClientBuilder();
                 builder.setClientCredentials(stream, iOSVariant.getPassphrase());
 
-                if (ProxyConfiguration.hasSocks()) {
-                    final Proxy proxy = ProxyConfiguration.socks();
-                    builder.setProxyHandlerFactory(new Socks5ProxyHandlerFactory(proxy.address()));
+                if (ProxyConfiguration.hasProxyConfig()) {
```
matzew: @PhilipGough with `hasProxyConfig` it sounds like a more generic question about proxy, but looking at the block of code below, I think it is really about `hasHttpProxyConfig`, right ?

service/src/main/java/org/jboss/aerogear/unifiedpush/service/impl/health/Ping.java
```diff
@@ -30,7 +31,12 @@ public static boolean isReachable(String host, int port) {
         try {
 
             if (ProxyConfiguration.hasSocks()) {
-                socket = new Socket(ProxyConfiguration.socks());
+                Proxy socksProxy = new Proxy(Proxy.Type.SOCKS,new InetSocketAddress(ProxyConfiguration.socks().getAddress(), ProxyConfiguration.socks().getPort()));
+                socket = new Socket(socksProxy);
+            } else if (ProxyConfiguration.hasProxyConfig()) {
```
matzew: same here, as above

PR #828, opened 2017-05-08
Fixes UPS GUI broken links in studio

josemigallas:
This PR solves the ticket: [AGPUSH-15180](https://issues.jboss.org/browse/RHMAP-15180)

Same as #826 

#### Reason
In a previous PR (#804) the ngLink directive that was causing a double HTTP request was changed into a simple href property. This might have caused the current issue.

#### Solution
Revert the changes made in #804.

#### Collateral Damage
Reverting to #804 will re-open the double requests issue so this fix should be only temporal until a better solution is found.


matzew:
:+1:

PR #861, opened 2017-07-18
[AGPUSH-2095][ADD] Add endpoint to show UPS version and scm details

diogoalbuquerque:
- add endpoint in HealthCheck.java to return manifest's attributes
- add organization name in aerogear-unifiedpush-server pom
- add SCM-URL and SCM-TAG in unifiedpush-server-wildfly pom

### DIFF
--- a/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
+++ b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
@@ -22,14 +22,20 @@ import org.jboss.aerogear.unifiedpush.service.impl.health.HealthDetails;
 import org.jboss.aerogear.unifiedpush.service.impl.health.HealthStatus;
 
 import javax.inject.Inject;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.io.InputStream;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
 
 /**
  * A class to test 'health' of the server
@@ -76,6 +82,29 @@ public class HealthCheck {
     @Produces(MediaType.TEXT_PLAIN)
     public Response ping() {
         return Response.ok("OK").build();
+    }
+
+
+    /**
+     * Endpoint to verify scm and version details
+     *
+     * @return {@link Attributes}
+     * @statuscode 200 Successful response for your request with manifest's attributes
+     * @statuscode 500 Error response for your resquest
+     */
+    @GET
+    @Path("/details")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response manifestDetails(@Context HttpServletRequest request) {
+
+        ServletContext context = request.getSession().getServletContext();
+
+        try {
+            InputStream manifestStream = context.getResourceAsStream("/META-INF/MANIFEST.MF");
+            return Response.ok(new Manifest(manifestStream).getMainAttributes()).build();
+        } catch (Exception e) {
+            return Response.serverError().build();
+        }
     }
 
 }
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -32,6 +32,10 @@
     <name>AeroGear UnifiedPush Server</name>
     <url>http://aerogear.org/push</url>
 
+    <organization>
```
matzew: our parent inherits from Jboss parent, which the correct info:
https://github.com/jboss/jboss-parent-pom/blob/10/pom.xml#L60-L63

No need to override it here

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
```diff
@@ -78,4 +84,27 @@ public Response ping() {
...
+
+    /**
+     * Endpoint to verify scm and version details
+     *
+     * @return {@link Attributes}
+     * @statuscode 200 Successful response for your request with manifest's attributes
+     * @statuscode 500 Error response for your resquest
+     */
+    @GET
+    @Path("/details")
```
matzew: how about `/version` ? 

@jasonmadigan, @ziccardi  any thoughts ?

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
```diff
@@ -78,4 +84,27 @@ public Response ping() {
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response manifestDetails(@Context HttpServletRequest request) {
+
+        ServletContext context = request.getSession().getServletContext();
+
+        try {
+            InputStream manifestStream = context.getResourceAsStream("/META-INF/MANIFEST.MF");
+            return Response.ok(new Manifest(manifestStream).getMainAttributes()).build();
+        } catch (Exception e) {
+            return Response.serverError().build();
```
matzew: let's not bubble that up - return 404

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
```diff
@@ -78,4 +84,27 @@ public Response ping() {
...
+     * @return {@link Attributes}
+     * @statuscode 200 Successful response for your request with manifest's attributes
+     * @statuscode 500 Error response for your resquest
+     */
+    @GET
+    @Path("/details")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response manifestDetails(@Context HttpServletRequest request) {
+
+        ServletContext context = request.getSession().getServletContext();
```
matzew: use `final` for `ServletContext` and `InputStream`

matzew:
Thanks for the PR! Sorry it took so long until we did review it

matzew:
@diogoalbuquerque thanks mate!

dimitraz:
Looks good 👍

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
```diff
@@ -78,4 +84,27 @@ public Response ping() {
...
+     */
+    @GET
+    @Path("/version")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response manifestDetails(@Context HttpServletRequest request) {
+
+        final ServletContext context = request.getSession().getServletContext();
+
+        try {
+            final InputStream manifestStream = context.getResourceAsStream("/META-INF/MANIFEST.MF");
```
ziccardi: As a rule of thumb, it would be better to always close streams.
I would suggest using 
`try(InputStream manifestStream = context.getResourceAsStream("/META-INF/MANIFEST.MF"))`

ziccardi:
@diogoalbuquerque thank you!

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/util/HealthCheck.java
matzew: +1

matzew:
I've merged your code, @diogoalbuquerque and also added the `try with resources` as @ziccardi was suggestion.

Thanks for your patient, and your help on this! Much appreciated! :tada: :balloon:

diogoalbuquerque:
@matzew No problem, I'm sorry for the delay in fix the try. 
You can always count on me, I'm very happy to contribute with this project. 👍
I will check more issues and I will work on them.

PR #867, opened 2017-07-26
[AGPUSH-2155] Check id before parsing the page size.

polinankoleva:
No need to parse pageSize when the id is null. Therefore make the check for id first then parsing of page size if needed.


matzew:
@polinankoleva makes sense moving it up. merged :tada:

PR #871, opened 2017-08-01
Lifeclycle APIs in JSR 236 are disabled, throwing IllegalStateExceptions

matzew:
@polinankoleva @dimitraz the lifecycle APIs of Managed ExecutorService in Java EE (JSR 236) are disabled, throwing a `IllegalStateException`


dimitraz:
Tested, no longer getting the exceptions 👍

PR #878, opened 2017-08-08
Reflecting HornetQ server settings

matzew:


matzew:
@josemigallas this should do the trick

rachael-oregan:
Need to remove ` max-redelivery-delay=5000` and `redelivery-delay=1.5`

matzew:
@rachael-oregan ah, that's likely on EAP 6.4.15 some less lenient xml rules :-) I used .11 where the file above worked. :rofl:

PR #881, opened 2017-08-09
Remove unused code from TokenLoader

matzew:
Some events are not used, so removing them.

@polinankoleva  @dimitraz


push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
```diff
@@ -31,7 +31,6 @@
 import javax.inject.Inject;
 import javax.jms.JMSException;
```
polinankoleva: BatchLoadedEvent and AllBatchesLoadedEvent imports can be removed.

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
```diff
@@ -107,7 +98,6 @@ public void loadAndQueueTokenBatch(@Observes @Dequeue MessageHolderWithVariants
         final Collection<Variant> variants = msg.getVariants();
```
polinankoleva: Update comment of the method. This statement "Additionally it fires {@link BatchLoadedEvent} as CDI event (that is translated to JMS event)."  is not valid anymore. The same for "When all batches were loaded for the given variant, it fires  {@link AllBatchesLoadedEvent}."

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
```diff
@@ -75,14 +74,6 @@
     @DispatchToQueue
     private Event<MessageHolderWithVariants> nextBatchEvent;
 
-    @Inject
-    @DispatchToQueue
-    private Event<BatchLoadedEvent> batchLoaded;
```
polinankoleva: If the events are not used anymore in the project, can we delete the classes BatchLoadedEvent and AllBatchesLoadedEvent?

push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/token/TokenLoader.java
matzew: Ah... the damn `@link` did fool me - otherwise, with unused import, we have checkstyle fails ;-) 

will change - good catch, thanks!

matzew:
@polinankoleva fixed

polinankoleva:
LGTM :) ready for merging

PR #891, opened 2017-08-16
AGPUSH-2052 - update router to prevent tab transition request executi…

rachael-oregan:
…ng twice

## Motivation
Angular UI executes requests twice when transitioning from one UI tab to another

## Result
This is a bug with the Angular router being used `angular-new-router` [here](https://github.com/angular/router/issues/204).  Stopped watching the `$location.path()` and updated the `Activity` and `Analytics` tab controllers.  As a result of not watching `$location.path()` the path in the url does not trigger a navigate request and therefore the back/forward buttons in the browsers do not work.

### DIFF
--- a/admin-ui/app/components/app-detail/include/activity.js
+++ b/admin-ui/app/components/app-detail/include/activity.js
@@ -97,9 +97,14 @@ angular.module('upsConsole')
       }
     });
 
-    $scope.$watch(function() { return self.searchString }, function( searchString ) {
-      self.currentPage = 1;
-      fetchMetrics( self.currentPage, self.searchString );
+    $scope.$watch(function() { 
+      return self.searchString 
+    }, function (newSearchString, oldSearchString) {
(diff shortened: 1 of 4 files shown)


matzew:
:eyes: at it

matzew:
I think it mostly works :joy: 

I get it fine for these TABs:
* Anaylics
* Variants
* Sender API

But when going to `Activity Log`, I get some XHRs twice:

e.g. against this URL `http://localhost:8080/ag-push/rest/metrics/messages/application/74933689-a061-4734-8585-66ac12ba6a3d?page=0&per_page=10&sort=desc`

rachael-oregan:
@matzew oh ya sorry should have mentioned that, the Activity Log calls the metrics endpoint in the controller again to ensure that the metrics get fetched.  When I removed this extra call to the endpoint I sometimes seen that the html page for that tab would not load. So I left in that extra call in there to ensure it works

matzew:
> So I left in that extra call in there to ensure it works

+1

PR #895, opened 2017-08-16
[AGPUSH-2176] Test evn setup for kafka consumer/producer

polinankoleva:


kafka/src/test/java/org/jboss/aerogear/unifiedpush/kafka/BasicKafkaTest.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public class BasicKafkaTest {
+
+    protected KafkaClusterWrapper kafkaCluster = new KafkaClusterWrapper();
+
+    /**
+     * Set system variables.
+     */
+    public static void init(){
+        System.setProperty("KAFKA_HOST", KafkaClusterWrapper.KAFKA_HOST);
```
matzew: Does it not work w/ the settings in the pom ? 

Not sure I really like this static "hook" into the Shrinkwrap deployment API

kafka/src/test/java/org/jboss/aerogear/unifiedpush/kafka/KafkaClusterWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.kafka.clients.consumer.OffsetResetStrategy;
+import io.debezium.kafka.KafkaCluster;
+import io.debezium.util.Testing;
+
+/**
+ * Kafka cluster wrapper class that adds useful methods for using it in the test environment.
+ */
+public class KafkaClusterWrapper {
+
+    public static final String KAFKA_HOST = "localhost";
```
matzew: I'd think we wanna use than the systems properties

kafka/src/test/java/org/jboss/aerogear/unifiedpush/kafka/consumers/InstallationMetricsConsumerTest.java
```diff
@@ -0,0 +1,109 @@
...
+                .addClass(InstallationMetricsKafkaConsumer.class)
+                // add a container for all mock providers
+                .addPackage(MockProviders.class.getPackage())
+                // needed to the cdi library
+                .addClass(DelegationKafkaConsumer.class).addClass(InstallationRegistrationEndpoint.class)
+                .addClass(AbstractBaseEndpoint.class).addAsManifestResource(EmptyAsset.INSTANCE, "beans.xml");
+    }
+
+    @Test
+    public void consumeTest() throws InterruptedException {
```
dimitraz: This test technically checks that the `updateAnalytics` method is called as expected ~ while that _should_ tell us that the message was consumed correctly, that's not necessarily the case so it might be nice to have another test which confirms this, similar to the one in the producer test?

dimitraz:
Good job :)

kafka/src/test/java/org/jboss/aerogear/unifiedpush/kafka/consumers/InstallationMetricsConsumerTest.java
polinankoleva: This tests checks not only if updateAnalytics is invoked. It checks if updateAnalytics is called with the exact parameter that is sent by the producer. And that check is done after "consume" method was called.

dimitraz: good point 😛

PR #899, opened 2017-08-18
Include server setup needed for live reload to work

rachael-oregan:


rachael-oregan:
@matzew @josemigallas @pb82 @sedroche

PR #910, opened 2017-08-25
moving push model classes into the pure model/api module

matzew:
the push model is really just a simple API, almost no dependencies.
Moving to the "model layer" will allow us a more fine grain approach, e.g. moving all models out to it's own repo... (e.g. some of these JARs are already used on other projects, like the java-sender)


matzew:
@ziccardi mind taking a look at this move ?

ziccardi:
+1

PR #913, opened 2017-08-28
[AGPUSH-2202] Update ReadMe after integration of Kafka

polinankoleva:
No need of Wildfly Full profile after removal of JMS.


matzew:
yay! :tada: 

thanks for this!

PR #751, opened 2016-09-08
[WIP] Keycloak and UPS separation 

sebastienblanc:
This is an updated version of https://github.com/aerogear/aerogear-unifiedpush-server/pull/677

Update since previous PR : 
- Keycloak and adapters (WF adapter and JS lib)  has been updated to 2.1.0
- Update to cli script
- Testing happen on WF 10.1.0
- Fix logout on UI

TODO : 
- [x] Test Scenario 1 
- [ ] Package AeroGear theme
- [ ] Update Docker image

c/p from previous PR : 

Changelog:
- Inclusion of Keycloak as bower component
  - Inclusion of usage instructions
  - Setup scripts
    - Shell script for Keycloak setup
    - Enable UPS to be used against whatever realm we want
    - Script to link UPS to Keycloak
  - Inclusion of configuration endpoint
    - Make it available for admin-ui
    - Enable UPS to be used against whatever realm we want
    - Fixes issues with hard coded references
  - Inclusion of Realm configuration template files
## Usage instructions

The following scenarios were tested against WildFly 9 and Keycloak 1.7.0.Final/1.8.0.CR3
# Scenario 1
### Description

The following setup covers scenarios where people wants to deploy Keycloak and UPS in separated Wildfly instances.
### Keycloak setup
1. Deploy Keycloak following [these instructions](http://keycloak.github.io/docs/userguide/keycloak-server/html/server-installation.html)
2. Import the realm configuration file and start WildFly/KeyCloak Server in a separate port
   `
   $UPS_HOME/bin/keycloak-setup.sh --ups-host=http://localhost:8082 --wildfly-offset=3 --realm-import=$UPS_HOME/bin/ups-realm-template.json
   `
3. Keycloak server should start with `aerogear` realm imported
### UPS setup
1. Follow the setup instructions described [here](https://aerogear.org/docs/unifiedpush/ups_userguide/index/#server-installation).
2. Setup the Keycloak adapter

```
  $UPS_HOME/bin/ups-setup.sh --adapter-setup
```
1. Start WildFly in a separate port
   `
   $WILDFLY_HOME/bin/standalone.sh -Djboss.socket.binding.port-offset=2 --server-config=standalone-full.xml
   `
2. Setup UPS instance to properly be attached to Keycloak instance
   
   ```
   $UPS_HOME/bin/ups-setup.sh --ups-host=localhost:9992 --realm=aerogear --auth-server=http://localhost:8083
   ```
3. Deploy the UnifiedPush server
# Scenario 2
[...]

PR #915, opened 2017-08-29
KC Decoupling to old 1.7.0

matzew:
replaces https://github.com/aerogear/aerogear-unifiedpush-server/pull/751  

## How to test
- Configure/run Keycloak
  - Easiest is: run the `keycloak-server.yaml` docker-compose file for Keycloak
  - check the IP address of the keycloak instance (e.g. via `docker inspect`)
  - login to Keycloak (`admin:admin`) and import the _sample_ realm (located in `keycloak/ups-realm-sample.json), which assumes that your WildFly for the UPS is on `localhost` and `8080`
- Run the AG-PUSH WAR file
  - Configure WF 10.1 (if not yet done), like [here](https://aerogear.org/docs/unifiedpush/ups_userguide/index/#server-installation)
  - in your WildFly 10.1, remove the current `auth-server` deployment (if present)
  - start the WF server with the following system properties: `-Dups.realm.name=aerogear -Dups.auth.server.url=http://IP_ADDRESS_OF_KEYCLOAK:8080/auth`
  - copy the newly build `ag-push.war` file to WF's deployment folder

## Expected:
* Login is redirected to the `IP OF KEYCLOAK`
* After successful UPS login, the UI comes up as normal


matzew:
@ziccardi mind reviewining / testing ?

matzew:
@pb82 can you also review and take a look ?

pb82:
@matzew Verified locally, thanks for the help!

PR #920, opened 2017-09-01
AGPUSH-2201 remove support from MPNS and SimplePush

pb82:
Remove support for MPSN and SimplePush from the server. Tests that still relied on one of those variants were modified to use different ones.


dependencies/server-dependencies/pom.xml
```diff
@@ -169,7 +169,7 @@
         <dependency>
             <groupId>org.aerogear.unifiedpush</groupId>
             <artifactId>admin-ui</artifactId>
-            <version>1.2.4</version>
+            <version>1.3.0</version>
```
matzew: once the JAR from the admin UI is release, this will be 1.2.6

until released, please use the snapshot.

We should never build non-snapshot things, outside of a release

matzew:
Has been merged to master: https://github.com/aerogear/aerogear-unifiedpush-server/commit/d747cd9d3d804fd411f448e95bd8467f42edfb8f 

@pb82 I did release the new UI, and was updating its usage here:
https://github.com/aerogear/aerogear-unifiedpush-server/blob/d747cd9d3d804fd411f448e95bd8467f42edfb8f/dependencies/server-dependencies/pom.xml#L166

PR #923, opened 2017-09-05
Adding dynamic keycloak origin

matzew:
For now using WildCards, instead of hard-coded URLs in our demo setup realm.

Note, for real production, this should be not wildcard


sebastienblanc:
:+1:

PR #925, opened 2017-09-06
adding tables to mirgator

matzew:
We forgot to add the new tables to the migrator script


matzew:
For now I wanna keep the "old" metrics tables, .. some users still might find it interesting/important 

but the migration needs to create these two new tables

PR #929, opened 2017-09-12
Replaced merge with getReference in delete method

ziccardi:
# Motivation
Currently the delete method used the merge approach to load the object to be deleted.
It has the drawback of loading the whole object from the database, while a simple reference would be enough.

# Solution
Using getReference instead of merge


ziccardi:
@matzew can you please review?

matzew:
@ziccardi can you port that to the 1.1.x-dev branch too ? 

Thanks!

PR #935, opened 2017-09-18
AGPUSH-1023 - Improve database .cli scripts for better user experience

rachael-oregan:
## Motivation
Currently the .cli scripts are very simple and don't handle exceptions well.  For example, when a duplicate resource error is thrown it is not handled nor is there any logging to give the user an explanation on how to fix this issue.

## Result
Add try/catch to the scripts to handle exceptions and errors. Extra logging is not needed as the scripts are able to handle the errors now and remove and add configurations as needed.  User's should be able to re-run these scripts when a config needs to be updated.

## Jira
https://issues.jboss.org/browse/AGPUSH-1023


rachael-oregan:
@matzew @ziccardi @josemigallas @kpiwko mind taking a look please. Thanks

matzew:
:high_brightness:   cool, thanks :tada: 
tested by executing the `prepare` script twice against a _fresh_ instance of WF11

PR #938, opened 2017-10-05
Bumps MySQL connector version to 5.1.44

josemigallas:

PR #939, opened 2017-10-05
Bumps MySQL connector version to 5.1.44

josemigallas:
This PR is the same as #938 but for master branch.


matzew:
Da script! :)

PR #942, opened 2017-11-27
KC 3.4.0 update

matzew:
prep work for UPS going KC 3.4.0


matzew:
@lgriffin I had to temporary disable the sonoar bits - looks like the build takes too long and/or fails just fails on Travis .

Once I have cut out the 1.2.1 release of UPS, I am fine w/ enabling it again, and we can case the issue - ok ?

PR #953, opened 2017-12-22
Latest KC

matzew:
parent updated to consume KC 3.4.2.Final release


matzew:
@philbrookes mind taking a quick look ? It's just dependency parent file update

PR #955, opened 2018-01-11
Fix and upgrade of OpenShift template

lbroudoux:
Hi, 

I've had to deploy UPS in production on OpenShift Online lately, and provided template was not "production-ready" (was not using Secret) and actually a little buggy.

I've patched the provided one by:
* adding secret management for provided templates
* removing Keycloak parameters that are now in dedicated external deployment

I have also added a new template embedding Keycloak for convenience.

This PR is related to https://github.com/aerogear/dockerfiles/pull/39 that should be applied to produce a Docker image able to reference a Keycloak instance running into OpenShift and reached through a Route (though on default http and https ports).

### DIFF
--- a/openshift/ups-keycloak-template.yaml
+++ b/openshift/ups-keycloak-template.yaml
@@ -1,7 +1,7 @@
 kind: "Template"
 apiVersion: "v1"
 metadata:
-  name: "aerogear-unified-push-wit-keycloak"
+  name: "aerogear-unified-push-with-keycloak"
   annotations:
     description: "The AeroGear Unified Push Server provides an API and web console for sending push messages to Android, iOS, Windows Phone, and other push messaging services. This templates embeds the deployment of a Keycloak server."
     tags: mobile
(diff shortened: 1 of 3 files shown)


openshift/ups-keycloak-template.yaml
```diff
@@ -0,0 +1,555 @@
...
+    required: true
+  - name: KEYCLOAK_IMAGE
+    displayName: Keycloak Image
+    description: Docker image to use for the Keycloak Server
+    value: jboss/keycloak-openshift:3.4.0.Final
+    required: true
+  - name: UPS_IMAGE
+    displayName: UPS Image
+    description: Docker image to use for the UPS Server
+    value: aerogear/unifiedpush-wildfly:1.2.1.3
```
matzew: can we use 1.2.3 ?

openshift/ups-keycloak-template.yaml
```diff
@@ -0,0 +1,555 @@
...
+    required: true
+  - name: MYSQL_IMAGE
+    displayName: MySQL Image
+    description: Docker image to use for the MySQL database servers
+    value: centos/mysql-57-centos7
+    required: true
+  - name: KEYCLOAK_IMAGE
+    displayName: Keycloak Image
+    description: Docker image to use for the Keycloak Server
+    value: jboss/keycloak-openshift:3.4.0.Final
```
matzew: can we use 3.4.3.Final ?

matzew:
Nice PR - thanks!

mind adding a little readme ?

lbroudoux:
Hi, I'm testing with the versions you mention and will add a little README after that.

matzew:
@lbroudoux hrm looks like your PR is a bit out of sync - mind to do a rebase ?

lbroudoux:
@matzew it should be ok now

matzew:
Thanks a lot, @lbroudoux !! :fireworks:

PR #960, opened 2018-01-30
using k8s compliant env vars

matzew:


dimitraz:
@matzew should this be updated as well?
https://github.com/aerogear/aerogear-unifiedpush-server/blob/Kafka_2018_rebase/kafka/src/test/java/org/jboss/aerogear/unifiedpush/kafka/KafkaClusterWrapper.java#L46

matzew:
@dimitraz good find, will update test and pom for the system props

PR #967, opened 2018-02-27
latest Prometheus client

matzew:
Updating to latest Prometheus client, containing some extra JVM info, see https://github.com/prometheus/client_java/pull/338  :smile_cat:


ziccardi:
@matzew Wouldn't be better to use variables instead of hardcoding the version?

matzew:
@ziccardi this is a property already, on the base pom. we use two different JARs, those are using the properties

ziccardi:
@matzew 👍

PR #973, opened 2018-03-23
Adding auto generate license file from WAR file

matzew:


matzew:
@ciaranRoche / @dimitraz if you are happy w/ the location, let's get this merged!

matzew:
@lgriffin feel free to review too :-)

lgriffin:
This looks good and is a great starting point certainly for the work in front of us. I'd suggest merging this as a good system of record. I do have some suggestions to have us more compliant with the wider MW teams which would be a good follow on PR:

1. Use of the canonical names: https://github.com/bucharest-gold/license-reporter/blob/master/lib/utils/resources/default-canonical-names.json

2. Use of the official metadata: https://github.com/bucharest-gold/license-reporter/blob/master/lib/utils/resources/default-unified-list.json e.g. official URL for the remote URL and also the approved flag. If something is NOT approved it should be a red flag :)

3. Add a distribution tag per license. This seems to be what EAP do for their Java based XML file.

```
<license>
    <name>Eclipse Public License, Version 1.0</name>
    <url>http://repository.jboss.org/licenses/epl-1.0.txt</url>
    <distribution>repo</distribution>
</license>
```
Everything else structurally looks perfect and this almost matches the official EAP XML structure save for the 3 suggestions above :)

matzew:
I think there is some MW plugin that might actually address 1) and 2) - I can follow up on that one, but when I started this (September) not much was there, to actually generate the lib.

> Add a distribution tag per license.

what's the valid/allowed values?   The generated XML file can be easily tweaked via a little plugin I wrote: https://github.com/aerogear/aerogear-jar-resource-bundle/blob/master/src/main/resources/META-INF/licenses.xml.vm

lgriffin:
All of the valid licenses are in the files I linked. All the metadata we default to is there as well. Ping me off this if I'm not answering your question :P

matzew:
> This looks good and is a great starting point certainly for the work in front of us

going to rebase and merge this

matzew:
@lgriffin @pb82 I've applied a few changes:
* we do not include our own JARs (they are composing the server), the `licenses.xml` just contains non aerogear dependencies
* We are not checking them in, since that is out of data, aspa. Instead we copy the SERVER licenses into the distribution folder
* adding LICENSE.txt to the distribtion as well

PR #974, opened 2018-04-01
Add psql stuff

matzew:
Adding installer CLI script, that setups up UPS/PSQL and adding docker-compose for PSQL, like we have for mysql


matzew:
was merged to master already - no need for any kind of need for that

PR #978, opened 2018-04-12
Allow Search Manager to work in different security contexts/environments

matzew:
* re-adding KC APIs to program against them, for env. checks
* refactoring: Moving `HttpBasicHelper` to our `common` module
* changing SearchManager to work in different envs

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -34,7 +34,11 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
-
+        <dependency>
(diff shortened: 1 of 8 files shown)


pb82:
@matzew i'm seeing this in the logs, is that ok?

```
09:17:48,139 ERROR [org.jboss.aerogear.unifiedpush.service.impl.SearchManager] (default task-2) The admin role: false
09:17:48,139 ERROR [org.jboss.aerogear.unifiedpush.service.impl.SearchManager] (default task-2) The developer role: false
```

PR #980, opened 2018-04-13
Adding docker compose for kc protected image

matzew:
Adding `dmp` to build a `aerogear/ups:kc` development image.

Run the build (e.g. `mvn clean install -DskipTests`, on ROOT folder)

Now, run the compose file:
```
cd servers/keycloak
mvn docker:run -Pcompose 
```
this brings up both, KC and UPS "dev" containers.

Now, go to `http://localhost:18080/ag-push`, and login (admin:123), change the passwd...

Now, create a App, and see something like
```
unifiedpushserver> 09:51:18,680 DEBUG [org.jboss.aerogear.unifiedpush.service.impl.SearchManager] (default task-2) Running with Keycloak context

```
in the log

### DIFF
--- a/servers/keycloak/src/main/docker/helper/exportKeycloakHost.sh
+++ b/servers/keycloak/src/main/docker/helper/exportKeycloakHost.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+## ping kecyloak host and extract IP address, to be used in KEYCLOAK_SERVICE_URL
+## this is needed since the UPS docker host needs to talk to the actual KC docker host,
+## hence the actual IP is needed
+export KEYCLOAK_SERVICE_HOST=$(ping -c 1 keycloak | grep -Eo -m 1 '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}';)
+echo KEYCLOAK runs on: $KEYCLOAK_SERVICE_HOST
+
+export KEYCLOAK_SERVICE_URL=http://$KEYCLOAK_SERVICE_HOST:8080/auth
+
+sh /opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0
--- a/servers/keycloak/src/main/docker/keycloak-realm/ups-realm-sample.json
+++ b/servers/keycloak/src/main/docker/keycloak-realm/ups-realm-sample.json
@@ -1,0 +1,69 @@
+{
+    "realm": "aerogear",
+    "enabled": true,
+    "accessTokenLifespan": 60,
+    "accessCodeLifespan": 60,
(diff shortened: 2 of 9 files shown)


servers/keycloak/src/main/docker/keycloak-realm/ups-realm-sample.json
```diff
@@ -0,0 +1,69 @@
...
+    "updateProfileOnInitialSocialLogin": false,
+    "requiredCredentials": [ "password" ],
+    "defaultRoles": [ "developer" ],
+    "users" : [
+        {
+            "username" : "admin",
+            "enabled": true,
+            "credentials" : [
+                { "type" : "password",
+                    "value" : "123" }
```
wtrocki: [Question] Is that standard password and account for keycloak? Maybe we need to put some minimal documentation for this new change?

servers/keycloak/src/main/docker/helper/exportKeycloakHost.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+export KEYCLOAK_SERVICE_HOST=$(ping -c 1 keycloak | grep -Eo -m 1 '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}';)
```
wtrocki: [Optiona][Trivial] Add comment here what this regexp does. It's not clear to me when trying that out.

wtrocki:
Verifying changes.

wtrocki:
Very nice change. It will be cool to add documentation for developers about this, however it could be in separate PR. Verified.

servers/keycloak/src/main/docker/keycloak-realm/ups-realm-sample.json
matzew: no, this is the ups admin, and its initial default password. we have this since ages

matzew:
> Very nice change. It will be cool to add documentation for developers about this, however it could be in separate PR

Yes! I was planing to write some README about this image, and the `plain` image as well

pb82:
👀

pb82:
@matzew The containers start without errors, but `http://localhost:18080/auth` return me a 404

danielpassos:
@pb82 I had the same problem. It was an error on the PR description. You should use `http://localhost:18080/ag-push` instead

matzew:
@pb82 @wtrocki @danielpassos 

yo - I am closing this PR; I will install KC into the _same_ WF server for UPS, using their `Overlay` distribution.  This is not supported for production, but does not matter here, since this is really just for testing.  i will get something new, soon :smile:

danielpassos:
¯\_(ツ)_/¯

PR #982, opened 2018-04-16
cleaner README and removing unneeded docker-compose files

matzew:
Some doc updates


README.md
```diff
@@ -25,92 +25,35 @@ The _AeroGear UnifiedPush Server_ is a server that allows sending push notificat
...
-And in a second terminal:
-
-    ./bin/standalone.sh -c standalone-full-ha.xml -Djboss.node.name=node2 -Djboss.messaging.cluster.password=somepassword -Djboss.socket.binding.port-offset=200 -Djava.net.preferIPv4Stack=true
+The easiest way to get started is running our `plain` Linux container:
+```
+mvn clean install -DskipTests && docker run -p 18081:8080 -it aerogear/ups:plain 
+```
 
-Note: on OS X, you need to enable multicast first:
+than go to `http://localhost:18081/` to use the UPS.
```
pb82: `than` -> `then`

docker-compose/README.md
```diff
@@ -0,0 +1,43 @@
...
+docker run -p 18081:8080 -it aerogear/ups:plain 
+```
+
+### Keycloak UPS 
+
+During the build of the project the `aerogear/ups:kc` is build. The image is protected with Keycloak, and for that we have `docker-compose` file, that can be easily executed like:
+
+```
+cd servers/keycloak
+mvn docker:run -Pcompose
```
pb82: Is this still required? We can also just run `docker run -p 8080:8080 -it aerogear/ups:kc`

docker-compose/docker-compose-v2.1.yaml
```diff
@@ -4,7 +4,7 @@ services:
   unifiedpushserver:
     image: docker.io/aerogear/unifiedpush-wildfly:1.3.0
     volumes:
-       - ./helper:/ups-helper
+       - ./helper:/ups-helper:z
```
pb82: what does the `z` at the end mean?

matzew: SELinux, for safer mounts, on linux 
(this setup - as you now know, anyway works on Linux, unless you do some fancy networking in docker)

matzew:
@pb82 thanks for the read, I've update the docker-compse doc

PR #983, opened 2018-04-17
Deploy UPS as ROOT.war, to serve on root context (/)

matzew:


docker-compose/README.md
```diff
@@ -31,7 +31,7 @@ During the build of the project the `aerogear/ups:plain` is build. Afterwards, s
 docker run -p 18081:8080 -it aerogear/ups:plain 
 ```
 
-This brings the UPS containers. Now, go to `http://localhost:18080/ag-push` and you can use it right away!
+This brings the UPS containers. Now, go to `http://localhost:18081/` and you can use it right away!
```
pb82: Why not also use `8080`  here?

matzew: to keep them different ? :small_airplane:

PR #984, opened 2018-04-17
Unify admin ui

pb82:
Make the auth config endpoint return a flag that tells the UI if auth is enabled or not. Also change the endpoint to make it not keycloak-specific.

ping @cfoskin

**Don't merge this yet, after verification the admin-ui needs to be released and this PR updated.**


jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/config/KeycloakConfigurationEndpoint.java
```diff
@@ -59,11 +59,13 @@ public Response configurationFile() throws JsonProcessingException {
 
         private String realm = "aerogear";
         @JsonProperty("auth-server-url")
-        private String authServerUrl = "/auth";
```
matzew: we can still keep the KC defaults here, since that internal/nested cfg is basically a KC config

matzew: but that's ok if not :)

PR #985, opened 2018-04-17
Latest postgres driver in setup script

matzew:


pb82:
👀

pb82:
@matzew I started wildfly 11 and ran the script. No errors. Then i tried to deploy the plain war (i tried both, war file and directory) but always end up with this error:

```
"{\"WFLYCTL0080: Failed services\" => {\"jboss.deployment.unit.\\\"ag-push.war\\\".PARSE\" => \"WFLYSRV0153: Failed to process phase PARSE of deployment \\\"ag-push.war\\\"
    Caused by: org.jboss.msc.service.ServiceNotFoundException: Service service jboss.ejb.default-resource-adapter-name-service not found\"}}"%   
```

pb82:
@matzew ah, looks like i need to use standalone-full.xml to get past this one

PR #986, opened 2018-04-17
Catch corner case where two errors for one push job are appended

matzew:
On  1.st request to Android, if there is an error AND we have topics and tokens, they are written same time to DB

Hence we have twice the same PK value

There is no harm, just verbose log. Fix is to catch and log

### DIFF
--- a/service/src/main/java/org/jboss/aerogear/unifiedpush/service/metrics/PushMessageMetricsService.java
+++ b/service/src/main/java/org/jboss/aerogear/unifiedpush/service/metrics/PushMessageMetricsService.java
@@ -29,6 +29,8 @@ import org.jboss.aerogear.unifiedpush.dao.PageResult;
 import org.jboss.aerogear.unifiedpush.dto.MessageMetrics;
 import org.jboss.aerogear.unifiedpush.system.ConfigurationUtils;
 import org.jboss.aerogear.unifiedpush.utils.DateUtils;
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/jboss/aerogear/unifiedpush/service/metrics/PushMessageMetricsService.java
```diff
@@ -64,7 +68,11 @@ public void updatePushMessageInformation(FlatPushMessageInformation pushMessageI
     public void appendError(final FlatPushMessageInformation pushMessageInformation, final Variant variant, final String errorMessage) {
         final VariantErrorStatus ves = new VariantErrorStatus(pushMessageInformation, variant, errorMessage);
         pushMessageInformation.getErrors().add(ves);
-        flatPushMessageInformationDao.update(pushMessageInformation);
+        try {
+            flatPushMessageInformationDao.update(pushMessageInformation);
+        } catch (Exception e) {
+            logger.info("A VariantErrorStatus has been already added, for the given Variant in this push job");
```
ziccardi: We catch a generic exception here, so the reason for the failure could be different than duplicate key.
I would change this error message and I would add the exception as well.
Something like: 
```java
   logger.info("Failed to save pushMessageInformation: {}", e.getMessage());
   logger.debug("Details:", e);
```

PR #987, opened 2018-04-17
Adding some more metric counterz

matzew:
Adding some more push metrics, per platform/batch job

See:
https://issues.jboss.org/browse/AGPUSH-2225

### DIFF
--- a/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/AdmPushNotificationSender.java
+++ b/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/AdmPushNotificationSender.java
@@ -17,6 +17,7 @@
 package org.jboss.aerogear.unifiedpush.message.sender;
 
 
+import io.prometheus.client.Counter;
 import org.jboss.aerogear.adm.ADM;
 import org.jboss.aerogear.adm.AdmService;
 import org.jboss.aerogear.adm.PayloadBuilder;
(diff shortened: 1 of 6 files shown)


pb82:
👀

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/AdmPushNotificationSender.java
```diff
@@ -34,6 +35,11 @@
 public class AdmPushNotificationSender implements PushNotificationSender {
     private final Logger logger = LoggerFactory.getLogger(AdmPushNotificationSender.class);
 
+    private static final Counter promPrushRequestsAdm = Counter.build()
+            .name("aerogear_ups_push_requests_adm")
```
pb82: What does the `adm` here stand for? Admin? Is there a way to trigger a batch requests in the UI?

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/WNSPushNotificationSender.java
```diff
@@ -58,6 +59,11 @@
 
     private final Logger logger = LoggerFactory.getLogger(WNSPushNotificationSender.class);
 
+    private static final Counter promPrushRequestsWindows = Counter.build()
+            .name("aerogear_ups_push_requests_windows_wns")
+            .help("Total number of Android push batch requests.")
```
pb82: Android -> Windows

pb82:
@matzew I sent push messages to Android and iOS Variants but only the total counter (aerogear_ups_push_requests_total) was increased. The other counters don't appear in the metrics output (aerogear_ups_push_requests_ios, aerogear_ups_push_requests_android).

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/AdmPushNotificationSender.java
matzew: adm -> Amazon device messaging (the class is also named ADMpush sender)

matzew:
> I sent push messages to Android and iOS Variants but only the total counter...

did you register devices ? If not - see in the code, on empty token collection, it immediately returns ;-)

PR #989, opened 2018-04-18
Complete assembly

matzew:
Adding missing artifacts to the distribution archive

* keycloak protected war
* admin-ui source files

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -75,6 +75,12 @@
                         <configuration>
                             <artifactItems>
                                 <artifactItem>
+                                    <groupId>org.aerogear.unifiedpush</groupId>
+                                    <artifactId>admin-ui</artifactId>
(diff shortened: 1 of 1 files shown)


matzew:
@pb82 regarding docker updates

the `keycloak` WAR needs to be used here:

https://github.com/aerogear/dockerfiles/blob/master/wildfly/unifiedpush-wildfly/Dockerfile#L29

We also need an image (similar to the [no-auth](https://github.com/aerogear/dockerfiles/blob/1.3.1.1.no-auth.Final/wildfly/unifiedpush-wildfly/Dockerfile#L7-L15)) here

I'd think the image for both _flavors_ are pretty much the same, just different WAR from the DIST file 

(of course the images come now w/ Postgres setup (e.g. drivers))

pb82:
👀

PR #990, opened 2018-04-20
Latest UPS image and updating to Postgres

matzew:
Updating to `unifiedpush-wildfly:2.0.0` (which is KC protected) and using Postgres DB (now our new default)

**NOTE:** this works fine on Linux


matzew:
> NOTE: this works fine on Linux

@evanshortiss similar to the `dockerfiles` repo, perhaps you might like to tweak this PR, adding the `--net=container....` feature ?   I've noticed that Mac users have issues w/ the different docker hosts, if no network is created.  /cc @pb82

pb82:
👀

PR #994, opened 2018-04-24
Adding some loggers and a bit of updates/refactorings

matzew:
Some nicer things:
* logs
* renaming to FCM (from GCM)
* and some comments


matzew:
> Looks like you also removed some old GCM code?

more a renaming to *FCM*

PR #996, opened 2018-04-26
remove experimental ams support

pb82:
Removing experimental ADM support from the UPS Server. UI will be done in a different PR.

Verificataion steps:

1. Build the UI from this branch: https://github.com/aerogear/unifiedpush-admin-ui/pull/26 using `mvn clean install`
1. Build UPS from this branch using `mvn clean install`
1. Run the generated docker image `docker run -p 8080:8080 -it aerogear/ups:plain`
1. Open a browser at `localhost:8080`
1. Create a push application and add a variant. No ADM option should be selectable.


matzew:
We should also remove it from the UI, release the UI, and than use that updated JAR


aother than that -> looks good

matzew:
works fine !

PR #998, opened 2018-04-30
include env vars in property lookup

pb82:
Support environment variables in addition to system properties for configuration variables. Env var names will be converted to adhere naming conventions, e.g. `custom.aerogear.fcm.push.host` will be converted into `CUSTOM_AEROGEAR_FCM_PUSH_HOST` before trying to resolve the env var.

### DIFF
--- a/common/src/main/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtils.java
+++ b/common/src/main/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtils.java
@@ -66,7 +66,7 @@ public final class ConfigurationUtils {
      * @param key String System Property Key
      * @return String Key as environment variable
      */
-    private static String formatEnvironmentVariable(String key) {
+    public static String formatEnvironmentVariable(String key) {
         return key.toUpperCase().replaceAll("\\.", "_");
     }
 
--- a/common/src/main/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtils.java
+++ b/common/src/main/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtils.java
@@ -24,20 +24,7 @@ public final class ConfigurationUtils {
 
     private static final Logger logger = LoggerFactory.getLogger(ConfigurationUtils.class);
 
-    private ConfigurationUtils() {
-        // no-op
-    }
-
-    /**
-     * Try to retrieve a system property and returns null if SecurityManager blocks it.
-     *
-     * @param key Name of the system property to get the string for.
-     *
-     * @return the value of the System property
-     */
-    public static String tryGetProperty(String key) {
-        return tryGetProperty(key, null);
-    }
+    private ConfigurationUtils() { }
 
     /**
      * Try to retrieve a system property and returns the defaultValue if SecurityManager blocks it.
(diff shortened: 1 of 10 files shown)


common/src/main/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtils.java
```diff
@@ -24,65 +24,110 @@
...
      *
      * @return the value of the System property
      */
-    public static String tryGetProperty(String key) {
-        return tryGetProperty(key, null);
+    private static String tryGetProperty(String key, String defaultValue) {
+        try {
+            return System.getProperty(key, defaultValue);
+        } catch (SecurityException e) {
+            logger.error("Could not get value of property {} due to SecurityManager. Using default value.", key);
```
matzew: nice

common/src/test/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtilsTest.java
```diff
@@ -37,28 +37,28 @@ public void cleanupTestProperty() {
     @Test
     public void testExistingTryGetProperty(){
         System.setProperty(TEST_PROPERTY_NAME, "MyNiceValue");
-        assertThat(ConfigurationUtils.tryGetProperty(TEST_PROPERTY_NAME)).isEqualTo("MyNiceValue");
+        assertThat(ConfigurationUtils.tryGetGlobalProperty(TEST_PROPERTY_NAME)).isEqualTo("MyNiceValue");
```
matzew: This all only tests the existing System properties, from the test.

Let's see if we can actually test ENV vars, includinge the fancy "toUpperCase and replaceAll" code.

I think it's possible to configure ENV_VARs in the pom file (sure fire plugin), like here:
https://github.com/aerogear/kafka-cdi/blob/master/pom.xml#L65-L77

but instead `systemPropertyVariables` it needs to be `environmentVariables`

matzew:
@pb82 beside the comment on testing, I think it's OK to merge it, but please add some tests, as suggested.

I wonder if we can have a markdown doc, somewhere in _this_ repo that explains all the System proerties and their ENV VAR counterparts - similar to the documentation of the UPS ? 

(can be a separate PR)

common/src/test/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtilsTest.java
pb82: @matzew Yes definitely makes sense to add a test here, will do. Unfortunately envirnonment variables can't easily be set from Java so i will need some mocking.

pb82:
@matzew Good idea with a list of available configuration variables. Will add it to the README.

common/src/test/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtilsTest.java
matzew: @pb82 I think, in the POM, on the surefire plugin they can be set - just a a few - I can take a look too, if you want

common/src/test/java/org/jboss/aerogear/unifiedpush/system/ConfigurationUtilsTest.java
```diff
@@ -37,28 +37,39 @@ public void cleanupTestProperty() {
...
 
+    @Test
+    public void testEnvVarFormat() {
+        assertThat(ConfigurationUtils.formatEnvironmentVariable("custom.aerogear.apns.push.host"))
+          .isEqualTo("CUSTOM_AEROGEAR_APNS_PUSH_HOST");
+    }
+
+    @Test
+    public void testEnvVarLookup() {
+        assertThat(ConfigurationUtils.tryGetGlobalProperty("test.env.var"))
```
matzew: Add a test for `TEST_ENV_VAR` as well

PR #999, opened 2018-04-30
AEROGEAR-2659 - check if cert is valid, otherwise do not try to connect

matzew:
this is adds code to check if a provided CERT is valid.
Only if valid, the connection to APNs happens...

See [AEROGEAR-2659](https://issues.jboss.org/browse/AEROGEAR-2659) for more

### DIFF
--- a/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
+++ b/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
@@ -21,15 +21,10 @@ import org.slf4j.LoggerFactory;
 
 import javax.security.auth.x500.X500Principal;
 import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.security.KeyStore;
 import java.security.cert.CertificateExpiredException;
 import java.security.cert.CertificateNotYetValidException;
 import java.security.cert.X509Certificate;
-import java.util.Calendar;
 import java.util.Enumeration;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -85,20 +80,17 @@ public final class ApnsUtil {
                 final String alias = aliases.nextElement();
                 final X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);
 
-
                 try{
                     certificate.checkValidity();
                 } catch (CertificateExpiredException | CertificateNotYetValidException e) {
-                    LOGGER.error("Provided APNs .p12 file is not valid");
-
+                    LOGGER.error("Provided APNs .p12 file is expired or not yet valid");
                     return false;
                 }
             }
+            return true;
         } catch (Exception e) {
             LOGGER.error("Error parsing .p12 file content", e);
             return false; // garbage is also not valid
         }
-
-        return true; // yes, it's all good, man!.
     }
 }
--- a/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
+++ b/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
@@ -21,8 +21,15 @@ import org.slf4j.LoggerFactory;
 
 import javax.security.auth.x500.X500Principal;
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.security.KeyStore;
+import java.security.cert.CertificateExpiredException;
+import java.security.cert.CertificateNotYetValidException;
 import java.security.cert.X509Certificate;
+import java.util.Calendar;
 import java.util.Enumeration;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -65,5 +72,33 @@ public final class ApnsUtil {
         }
 
         return null; // if no topic was found we try with null
+    }
+
+    public static boolean checkValidity(final byte[] keystore, final char[] password) {
+        try {
+            final KeyStore keyStore = KeyStore.getInstance(KEYSTORE_TYPE);
+            keyStore.load(new ByteArrayInputStream(keystore), password);
+
+            final Enumeration<String> aliases = keyStore.aliases();
+
+            while (aliases.hasMoreElements()) {
+                final String alias = aliases.nextElement();
+                final X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);
+
+
+                try{
+                    certificate.checkValidity();
+                } catch (CertificateExpiredException | CertificateNotYetValidException e) {
+                    LOGGER.error("Provided APNs .p12 file is not valid");
+
+                    return false;
+                }
+            }
+        } catch (Exception e) {
+            LOGGER.error("Error parsing .p12 file content", e);
+            return false; // garbage is also not valid
+        }
+
+        return true; // yes, it's all good, man!.
     }
 }
--- a/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
+++ b/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
@@ -25,14 +25,22 @@ import java.security.KeyStore;
 import java.security.cert.CertificateExpiredException;
 import java.security.cert.CertificateNotYetValidException;
 import java.security.cert.X509Certificate;
+import java.util.Arrays;
 import java.util.Enumeration;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public final class ApnsUtil {
 
     private static final String KEYSTORE_TYPE = "PKCS12";
-    private static final String TOPIC_PATTERN = ".*UID=([^,]+).*";
+    private static final Pattern TOPIC_PATTERN = Pattern.compile(".*UID=([^,]+).*");
+    private static final Pattern COMMON_NAME_PATTERN = Pattern.compile("CN=(.*?)\\:");
+    private static final List<String> PUSH_SUBJECTS =
+            Arrays.asList(
+                    "Apple Push Services",
+                    "Apple Production IOS Push Services",
+                    "Apple Development IOS Push Services");
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ApnsUtil.class);
 
@@ -54,8 +62,7 @@ public final class ApnsUtil {
                 final String subject = subjectX500Principal.getName();
                 if (subject != null ) {
 
-                    final Pattern pattern = Pattern.compile(TOPIC_PATTERN);
-                    final Matcher matcher = pattern.matcher(subject);
+                    final Matcher matcher = TOPIC_PATTERN.matcher(subject);
 
                     if (matcher.matches()) {
                         return matcher.group(1);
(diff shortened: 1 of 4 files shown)


push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -66,4 +73,32 @@ public static String readDefaultTopic(final byte[] keystore, final char[] passwo
...
+
+            while (aliases.hasMoreElements()) {
+                final String alias = aliases.nextElement();
+                final X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);
+
+
+                try{
+                    certificate.checkValidity();
+                } catch (CertificateExpiredException | CertificateNotYetValidException e) {
+                    LOGGER.error("Provided APNs .p12 file is not valid");
```
matzew: with a lot of pushes, e.g. from hooks / automated schedulers etc, this might generate a lot of ERROR loggings...

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -66,4 +73,32 @@ public static String readDefaultTopic(final byte[] keystore, final char[] passwo
...
+
+                    return false;
+                }
+            }
+        } catch (Exception e) {
+            LOGGER.error("Error parsing .p12 file content", e);
+            return false; // garbage is also not valid
+        }
+
+        return true; // yes, it's all good, man!.
```
ziccardi: this can be reduced to a `return true` right after the `checkValidity()` and a `return false` at the end of the method. I think the code is more clear with less exit paths.

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -66,4 +73,32 @@ public static String readDefaultTopic(final byte[] keystore, final char[] passwo
...
+
+            final Enumeration<String> aliases = keyStore.aliases();
+
+            while (aliases.hasMoreElements()) {
+                final String alias = aliases.nextElement();
+                final X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);
+
+
+                try{
+                    certificate.checkValidity();
```
ziccardi: A .p12 can contain many certificates, even certificates we don't care at all. Do we want to fail if _any_ certificate inside the p12 is expired?

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
```diff
@@ -208,35 +212,42 @@ private ApnsClient buildApnsClient(final iOSVariant iOSVariant) {
...
-                if (ProxyConfiguration.hasHttpProxyConfig()) {
-                    if (ProxyConfiguration.hasBasicAuth()) {
-                        String user =  ProxyConfiguration.getProxyUser();
-                        String pass = ProxyConfiguration.getProxyPass();
-                        builder.setProxyHandlerFactory(new HttpProxyHandlerFactory(ProxyConfiguration.proxyAddress(), user, pass));
-                    } else {
-                        builder.setProxyHandlerFactory(new HttpProxyHandlerFactory(ProxyConfiguration.proxyAddress()));
-                    }
+            // only process valid certs!
+            if (ApnsUtil.checkValidity(iOSVariant.getCertificate(), iOSVariant.getPassphrase().toCharArray())) {
```
pb82: since this is done in `buildApnsClient` it should only be checked once per cached client right? (So no check for every batch, which would be unneccessary)

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
```diff
@@ -208,35 +212,42 @@ private ApnsClient buildApnsClient(final iOSVariant iOSVariant) {
...
 
-                } else if (ProxyConfiguration.hasSocksProxyConfig()) {
-                    builder.setProxyHandlerFactory(new Socks5ProxyHandlerFactory(ProxyConfiguration.socks()));
-                }
+                // add the certificate:
+                try {
+                    final ByteArrayInputStream stream = new ByteArrayInputStream(iOSVariant.getCertificate());
```
ziccardi: we (correctly) close this stream even if it is a ByteArrayInputStream. Would be better to change this to a "try with resource"

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
```diff
@@ -208,35 +212,42 @@ private ApnsClient buildApnsClient(final iOSVariant iOSVariant) {
...
-                logger.error("Error reading certificate", e);
-                // will be thrown below
+                    } else if (ProxyConfiguration.hasSocksProxyConfig()) {
+                        builder.setProxyHandlerFactory(new Socks5ProxyHandlerFactory(ProxyConfiguration.socks()));
+                    }
+
+                    final ApnsClient apnsClient = builder.build();
+
+                    // release the stream
+                    stream.close();
```
ziccardi: Useless if we change to a "try with resource"

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
```diff
@@ -109,7 +109,7 @@ public void sendPushMessage(final Variant variant, final Collection<String> toke
             }
         }
 
-        if (apnsClient.isConnected()) {
+        if (apnsClient != null && apnsClient.isConnected()) {
```
pb82: why the null check? It looks like `receiveApnsConnection` can either return a valid client or throw.

pb82:
Tested with an old certificate and got:
```
14:31:58,109 ERROR [org.jboss.aerogear.unifiedpush.message.sender.apns.ApnsUtil] (Thread-1 (ActiveMQ-client-global-threads)) Provided APNs .p12 file is not valid
```
followed by a gigantic stack trace starting with:
```
14:31:58,111 ERROR [org.jboss.as.ejb3.invocation] (Thread-1 (ActiveMQ-client-global-threads)) WFLYEJB0034: EJB Invocation failed on component SimpleApnsClientCache for method public com.turo.pushy.apns.ApnsClient org.jboss.aerogear.unifiedpush.message.cache.SimpleApnsClientCache.getApnsClientForVariant(org.jboss.aerogear.unifiedpush.api.iOSVariant,org.jboss.aerogear.unifiedpush.message.cache.ServiceConstructor): javax.ejb.EJBTransactionRolledbackException
```

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
pb82: @ziccardi Hmm, i would assume so since all of them should be issued by Apple for accessing APNS right? Or do you think there could be unrelated certs in there as well?

ziccardi: @pb82 A P12 could contain any kind of certificate. I think we should at least check if the CN is related to sites we are interested in.

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/ApnsUtil.java
```diff
@@ -66,4 +68,29 @@ public static String readDefaultTopic(final byte[] keystore, final char[] passwo
...
+            final Enumeration<String> aliases = keyStore.aliases();
+
+            while (aliases.hasMoreElements()) {
+                final String alias = aliases.nextElement();
+                final X509Certificate certificate = (X509Certificate) keyStore.getCertificate(alias);
+
+                try{
+                    certificate.checkValidity();
+                } catch (CertificateExpiredException | CertificateNotYetValidException e) {
+                    LOGGER.error("Provided APNs .p12 file is expired or not yet valid");
```
matzew: So, this can - depending on the actual amount of push apptempts cause some mega logs, in case the customer/user has invalid push - and does not bother to update

matzew: But somehow, I feel this should be logged ... since it is an issue that requires action from the owner of the certificate

matzew: By default the log looks like:

```

14:52:21,237 DEBUG [org.jboss.aerogear.unifiedpush.message.cache.SimpleApnsClientCache] (Thread-1 (ActiveMQ-client-global-threads)) no cached connection for e0351314-c5a4-4b52-beef-e8e8626edb5c-dev, establishing it
14:52:21,242 ERROR [org.jboss.aerogear.unifiedpush.message.sender.apns.ApnsUtil] (Thread-1 (ActiveMQ-client-global-threads)) Provided APNs .p12 file is expired
14:52:21,242 ERROR [org.jboss.aerogear.unifiedpush.message.sender.apns.PushyApnsSender] (Thread-1 (ActiveMQ-client-global-threads)) Unable to send notifications, client is not connected. Removing from cache pool
14:52:21,242 WARN  [org.jboss.aerogear.unifiedpush.message.NotificationDispatcher] (Thread-1 (ActiveMQ-client-global-threads)) Error on 'ios' delivery: Unable to send notifications, client is not connected
14:52:21,251 DEBUG [org.jboss.aerogear.unifiedpush.message.cache.SimpleApnsClientCache] (Thread-1 (ActiveMQ-client-global-threads)) Removed client from cache for e0351314-c5a4-4b52-beef-e8e8626edb5c
```

for each invocation, w/ an invalid certificate

matzew: Of course `DEBUG` is only in our own Docker image enabled.

so, each would do 2 ERROR and one WARN

pb82: Would it be better to check the certificate in the NotificationDispatcher: https://github.com/aerogear/aerogear-unifiedpush-server/blob/master/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/NotificationDispatcher.java#L66 ?

But at this point we would have to check if we are dealing with APNS, then cast the sender and the variant and then do the check.

matzew:
@ziccardi can you comment regarding the CN (Common Name) of the .p12 file ? 

Otherwise I think the solution is perfect, and even shows the "invalid certificat" on the push server UI (activity tab)

ziccardi:
@matzew @pb82 The solution looks ok to me as well, but if we don't check the CN of the certificate (which should contain the url of the site to be authenticated), then we will give an error even if a totally unrelated certificate is expired or untrusted and I think it could generate confusion and would be a bug.

matzew:
Ok, fair point, @ziccardi 

I have two .p12 files here, one is mine (DEVELOPMENT TYPE) and one from @pb82 (not sure if production or DEV)

When I have this printed:

```
final X500Principal subjectX500Principal = certificate.getSubjectX500Principal();
final String subject = subjectX500Principal.getName();
```

The subject returns a longer string, including the `CN`.  Here is the output, where I only have the pasted the `CN` part:

* `.....CN=Apple Development IOS Push Services: THE_APP_ID.....`
* `.....CN=Apple Push Services: THE_APP_ID.....`

Should we check if there is something (e.g. contains, or fancy regex), like "Apple *** Push Services" in the `CN` section ? 

@ziccardi feel free to paste some code here, that @pb82 can apply later / tomorrow

matzew:
@pb82 @ziccardi I found an interesting POSTs, here: https://github.com/parse-community/Parse-SDK-iOS-OSX/issues/672#issue-122944350 and here: https://stackoverflow.com/questions/34349849/ios-push-notificattion-production-certificate-common-name-issue

Indeeded, my CERT is pretty old, and just updated, hence the `Apple Development IOS Push Services`

To be safe, I think we want those as possible options on the `contains()` check?
* Apple Production IOS Push Services
* Apple Development IOS Push Services
* Apple Push Services

PR #1002, opened 2018-05-09
fix: Check variant platform before deletion/secretReset/findById

aliok:
JIRA: https://issues.jboss.org/browse/AGDROID-810

UPS offers a REST API to delete variants in the form of 'https://SERVER:PORT/CONTEXT/rest/applications/pushAppID/<PLATFORM>/variantId/'.

However the <PLATFORM> part (android or ios) is not checked and you can use either either value to delete a variant of another kind. This is not a critical bug because you still need a valid variant ID but is should be fixed.

### Impl
Convert the endpoint implementations in the base class to common methods and use them in the variant endpoint implementations.

Commits:
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/dd8b7f5cd0062364bc8253b9fca27929d9b8c50c: Initial work 
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/c55c3c58505a901346ee79d218909a748a7c2f3d: Return 400 instead of 404 if found variant is of another type
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/2b709729c44cfab7a219e7cbc519f7aab3c54467: Tests for the modified parts above
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/a2bc938729cd1a1766e652b2d9c7b78036c5315b: Tests for other operations in the variant endpoints were missing, so I wrote them.
  * In order to not go crazy with tests, I only wrote base case tests for AndroidVariantEndpoint! IOS and Windows endpoints have tests for happy cases
  * Needed to convert some field injections to constructor injections for testability. I've used the strategy "mandatory shit in constructor injections, optionals in field injections"
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/20c9fb861e86606ca1a1fe6bde2c10706303bea6: RESTEasy (JAX WS impl we use) requires no-arg constructors on endpoints and won't work with the `@Inject` constuctors. So, needed to revert constructor injectors to field injectors. But, in order to make things testable, I added additional `protected` constructors to pass params to those endpoints. RESTEasy seems happy about that.
* https://github.com/aerogear/aerogear-unifiedpush-server/pull/1002/commits/317b453b985236fa6ca436c4f3c4cc2d2ce2bd95: Generify variant endpoint base class

### DIFF
--- a/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
+++ b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
@@ -23,14 +23,6 @@ import org.jboss.aerogear.unifiedpush.service.GenericVariantService;
 import org.jboss.aerogear.unifiedpush.service.PushApplicationService;
 
 import javax.inject.Inject;
-import javax.ws.rs.Consumes;
-import javax.ws.rs.DELETE;
-import javax.ws.rs.GET;
-import javax.ws.rs.PUT;
-import javax.ws.rs.Path;
-import javax.ws.rs.PathParam;
-import javax.ws.rs.Produces;
-import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.util.Objects;
 import java.util.Set;
@@ -56,28 +48,29 @@ public abstract class AbstractVariantEndpoint extends AbstractBaseEndpoint {
      * @return          {@link Variant} with new secret
      *
      * @statuscode 200 The secret of Variant reset successfully
-     * @statuscode 404 The requested Variant resource does not exist
+     * @statuscode 404 The requested Variant resource does not exist or it is not of the given type
      */
-    @PUT
-    @Path("/{variantId}/reset")
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces(MediaType.APPLICATION_JSON)
-    public Response resetSecret(@PathParam("variantId") String variantId) {
+    protected <T extends Variant> Response doResetSecret(String variantId, Class<T> type) {
 
         Variant variant = variantService.findByVariantID(variantId);
 
-        if (variant != null) {
-            logger.trace("Resetting secret for: {}", variant.getName());
-
-            // generate the new 'secret' and apply it:
-            String newSecret = UUID.randomUUID().toString();
-            variant.setSecret(newSecret);
-            variantService.updateVariant(variant);
-
-            return Response.ok(variant).build();
+        if (variant == null) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
         }
 
-        return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
+        if (!type.isInstance(variant)) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Requested Variant is of another type/platform").build();
+        }
+
+        logger.trace("Resetting secret for: {}", variant.getName());
+
+        // generate the new 'secret' and apply it:
+        String newSecret = UUID.randomUUID().toString();
+        variant.setSecret(newSecret);
+        variantService.updateVariant(variant);
+
+        return Response.ok(variant).build();
+
     }
 
     /**
@@ -86,20 +79,21 @@ public abstract class AbstractVariantEndpoint extends AbstractBaseEndpoint {
      * @param variantId id of {@link Variant}
      * @return          requested {@link Variant}
      *
-     * @statuscode 404 The requested Variant resource does not exist
+     * @statuscode 404 The requested Variant resource does not exist or it is not of the given type
      */
-    @GET
-    @Path("/{variantId}")
-    @Produces(MediaType.APPLICATION_JSON)
-    public Response findVariantById(@PathParam("variantId") String variantId) {
+    protected <T extends Variant> Response doFindVariantById(String variantId, Class<T> type) {
 
         Variant variant = variantService.findByVariantID(variantId);
 
-        if (variant != null) {
-            return Response.ok(variant).build();
+        if (variant == null) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
         }
 
-        return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
+        if (!type.isInstance(variant)) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Requested Variant is of another type/platform").build();
+        }
+
+        return Response.ok(variant).build();
     }
 
     /**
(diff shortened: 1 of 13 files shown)


aliok:
@pb82 went ahead and created the PR as WIP before going crazy with the tests.

@matzew we couldn't find any tests for these variant endpoints. wanted to make sure before spending lots of time creating them. no tests for these, right?

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
```diff
@@ -110,22 +104,24 @@ public Response findVariantById(@PathParam("variantId") String variantId) {
...
-        if (variant != null) {
-            logger.trace("Deleting: {}", variant.getClass().getSimpleName());
+        if (variant == null) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Could not find requested Variant").build();
+        }
 
-            variantService.removeVariant(variant);
-            return Response.noContent().build();
+        if (!type.isInstance(variant)) {
+            return Response.status(Response.Status.NOT_FOUND).entity("Requested Variant is of another type/platform").build();
```
pb82: Maybe better return `400 - Bad request` instead of 404 in this case

pb82: Applies to the other methods too

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
```diff
@@ -151,4 +143,52 @@ public Response updateWindowsVariation(
...
+     *
+     * @param variantId id of {@link Variant}
+     * @return          {@link Variant} with new secret
+     *
+     * @statuscode 200 The secret of Windows Variant reset successfully
+     * @statuscode 404 The requested Windows Variant resource does not exist
+     */
+    @PUT
+    @Path("/{variantId}/reset")
+    @Consumes(MediaType.APPLICATION_JSON)
```
pb82: Not really part of this fix but why is the `@Consume` annotation here? It looks like this method does not receive any JSON and only uses the variantId from the path?

pb82:
@aliok Looks good:

```                                                                                            
peter@Peters-MBP-2:~|⇒  curl -X DELETE http://localhost:8080/rest/applications/f250fa30-8bc7-4a28-8991-3a2833d7e4ba/ios/eb7f8fa9-ef75-4fc5-9bb9-a48b4864c769
Requested Variant is of another type/platform%                                                                                             peter@Peters-MBP-2:~|⇒  curl -X DELETE http://localhost:8080/rest/applications/f250fa30-8bc7-4a28-8991-3a2833d7e4ba/android/eb7f8fa9-ef75-4fc5-9bb9-a48b4864c769
peter@Peters-MBP-2:~|⇒  
```

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/WindowsVariantEndpoint.java
aliok: No idea. Just moved this method from the base class. Let me delete it.

aliok:
ready for review... please see the PR description for explanation of things.

aliok:
@pb82 I noticed that there are 2 update methods on IOSvariantEndpoint:
1. With multipart data: https://github.com/aliok/aerogear-unifiedpush-server/blob/a2bc938729cd1a1766e652b2d9c7b78036c5315b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/iOSVariantEndpoint.java#L184
2. With post data: https://github.com/aliok/aerogear-unifiedpush-server/blob/a2bc938729cd1a1766e652b2d9c7b78036c5315b/jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/iOSVariantEndpoint.java#L146

(2) is not doing the validations (1) does and it does not fire the event (1) fires. I think (2) is obsolete and should be removed. How can we confirm this?

aliok:
ah, and @pb82 mind doing the basic CURL tests you did?

aliok:
@psturc  pointed me that when he tries to start UPS, he got an exception. 
Because of the reasons written here: https://issues.jboss.org/browse/RESTEASY-1538, I need to revert the constructor injections to field injections as before.

aliok:
@psturc should be working now. I was able to start up UPS and create bunch of variants and apps from the UI.

pb82:
@aliok @psturc The easiest and best way to start UPS locally is using the docker images that get created automatically by maven. Just run:

```
$ mvn clean install
$ docker run -p 8080:8080 -it aerogear/ups:plain
```

This will start UPS without the need for an external database or Keycloak server.

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AndroidVariantEndpoint.java
```diff
@@ -147,4 +153,54 @@ public Response updateAndroidVariation(
...
+     * @return no content or 404
+     *
+     * @statuscode 204 The Variant successfully deleted
+     * @statuscode 400 The requested Variant resource exists but it is not for Android
+     * @statuscode 404 The requested Android Variant resource does not exist
+     */
+    @DELETE
+    @Path("/{variantId}")
+    public Response deleteVariant(@PathParam("variantId") String variantId) {
+        return this.doDeleteVariant(variantId, AndroidVariant.class);
```
ziccardi: I think using `this` here is confusing since the method is implemented in the superclass.

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
```diff
@@ -49,35 +43,46 @@
     @Inject
     protected GenericVariantService variantService;
 
+    // required for RESTEasy
+    public AbstractVariantEndpoint() {
```
ziccardi: If we would change this to
```java
public abstract class AbstractVariantEndpoint <T extends Variant> {
   private final Class<T> type;

   protected abstract AbstractVariantEndpoint(Class<T> type) {
   }
```

we could remove all the `doXXX` method and we won't need to implement a custom XXX method in each subclass.

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
```diff
@@ -49,35 +43,46 @@
...
      * @statuscode 200 The secret of Variant reset successfully
+     * @statuscode 400 The requested Variant resource exists but it is not of the given type
      * @statuscode 404 The requested Variant resource does not exist
      */
-    @PUT
-    @Path("/{variantId}/reset")
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces(MediaType.APPLICATION_JSON)
-    public Response resetSecret(@PathParam("variantId") String variantId) {
+    protected <T extends Variant> Response doResetSecret(String variantId, Class<T> type) {
```
ziccardi: changing the constructor as suggested before, this can be changed again to 
```java
public Response resetSecret(@PathParam("variantId") String variantId)
```

aliok:
@ziccardi good points. applied some changes as you suggested.
however, instead of having the type in the constructor, I added an abstract `getVariantType()` method that the subclasses implement. RESTEasy/JAX.RS doesn't like constructor params.

I verified my changes using the admin console, which uses the REST api.

ziccardi:
@aliok : the constructor param would be only in the base abstract class. The concrete implementation won't have any parameters, so that should not be an issue for RESTEasy

aliok:
@ziccardi you're right. amended my commit that way.
thanks for the feedback.

jaxrs/src/main/java/org/jboss/aerogear/unifiedpush/rest/registry/applications/AbstractVariantEndpoint.java
```diff
@@ -41,21 +36,36 @@
...
+
+    private final Class<T> type;
 
     @Inject
     protected PushApplicationService pushAppService;
 
     @Inject
     protected GenericVariantService variantService;
 
+    // required for RESTEasy
```
ziccardi: Remove this comment

PR #1003, opened 2018-05-09
Add support for mutable-content key

qultist:
The [UNNotificationServiceExtension](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) was introduced in iOS 10 which adds the possibility to modify the remote notification’s content before displaying it. But the extension gets called only if the mutable-content key’s value is set to 1.

Since the underlying pushy supports the key the changes are very small.


pb82:
@qultist  Hey, thanks for your contribution! Sorry for only getting to this now. Will take a look and perform some testing but this looks good!

PR #1004, opened 2018-05-26
better organization of the dmp and latest version

matzew:
Moved the `dmp` to the plugin management section of the servers artifact.

This allows to quick updates (e.g. version) for all nested projects


pb82:
👀

PR #1011, opened 2019-01-30
Add docker packaging to separate profiles

jomrazek:
## Motivation
In productization we need to run maven and image builds separate from each other, this option was missing from current version of ups.

## What
Added profile to deactivate docker assembly with maven property while leaving default behavior without changes.

## Checklist:

- [X] Code has been tested locally by PR requester

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/servers/keycloak/pom.xml
+++ b/servers/keycloak/pom.xml
@@ -15,7 +15,8 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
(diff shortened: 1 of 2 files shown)


pb82:
@jomrazek How do i correctly activate the profile? I tried it with `mvn clean install -Pnodocker` but the docker maven plugin is still invoked.

jomrazek:
@pb82 profile is active by default to keep your standard behaviour, by passing -Dnodocker=true you change it to not run.
`mvn clean install -Dnodocker=true`

wei-lee:
@secondsun do you mind take a look at this one and see if we can merge it as part of this release?

PR #1014, opened 2019-03-06
fix:connect to postgres in docker image

wei-lee:
https://issues.jboss.org/browse/AEROGEAR-8764

### DIFF
--- a/servers/plain/src/main/docker/Dockerfile
+++ b/servers/plain/src/main/docker/Dockerfile
@@ -1,16 +1,12 @@
-# Use latest jboss/wildfly image as the base
 FROM ${base}
 MAINTAINER AeroGear <aerogear@googlegroups.com>
 
-# Run everything below as root user
 USER root
 
-# Clean the metadata
 RUN yum install -y unzip wget && yum -q clean all
 
 ARG CACHEBUST=1
 
-## PostgreSQL
 ENV psql_module_dir=$JBOSS_HOME/modules/org/postgresql/main/
 ENV psql_connector_jar=postgresql-jdbc.jar
 RUN mkdir -p ${psql_module_dir}
@@ -19,16 +15,12 @@ RUN mv ${psql_connector_jar} ${psql_module_dir}
 
 ADD ${postgresConfigFile} ${psql_module_dir}/module.xml
 
-# WildFly configuration file ready for JMS and HSQL
 ADD ${configFile} $JBOSS_HOME/standalone/configuration/standalone.xml
 
-# Run everything below as aerogear user
 USER jboss
 
-# Switch to the working dir $JBOSS_HOME/standalone/deployments
 WORKDIR /opt/jboss/wildfly/standalone/deployments
 
-# add war files
 COPY maven/ $JBOSS_HOME/standalone/deployments
 
 COPY entrypoint.sh /opt/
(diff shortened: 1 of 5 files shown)


grdryn:
Suggestion: change the `MAINTAINER` instruction in the Dockerfile

wei-lee:
I have built an image for testing: https://hub.docker.com/r/weilee/ups/tags. It is deployed here: https://master.psturc-8098.openshiftworkshop.com/console/project/wei-test/overview.

To verify, create some data in UPS, redeploy and the data should not go away.

servers/plain/src/main/docker/Dockerfile
```diff
@@ -10,12 +10,18 @@ RUN yum install -y unzip wget && yum -q clean all
...
+
+ADD ${postgresConfigFile} ${psql_module_dir}/module.xml
+
 # WildFly configuration file ready for JMS and HSQL
 ADD ${configFile} $JBOSS_HOME/standalone/configuration/standalone.xml
 
-# Switch to the working dir /opt/jboss/wildfly
-WORKDIR /opt/jboss/wildfly
-
 # Run everything below as aerogear user
```
ziccardi: No part of this change, but I believe this should read `Run everything below as 'jboss' user`

grdryn: Agreed. A comment probably isn't needed for this even (a lot of the comments here could probably be removed)!

servers/plain/src/main/docker/Dockerfile
```diff
@@ -1,29 +1,27 @@
...
 
-# Switch to the working dir /opt/jboss/wildfly
-WORKDIR /opt/jboss/wildfly
+ADD ${configFile} $JBOSS_HOME/standalone/configuration/standalone.xml
 
-# Run everything below as aerogear user
 USER jboss
 
-# Switch to the working dir $JBOSS_HOME/standalone/deployments
 WORKDIR /opt/jboss/wildfly/standalone/deployments
```
grdryn: Everywhere else, this `/opt/jboss/wildfly` is specified using `$JBOSS_HOME` (see the next `COPY` for example). Could/should it be `$JBOSS_HOME` here too?

servers/plain/src/main/docker/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -e
+
+# launch wildfly
+echo "launching wildfly"
+exec /opt/jboss/wildfly/bin/standalone.sh -b 0.0.0.0 $@
```
grdryn: If we're going to keep this file (which I think is not adding much value at this point), the path to the script should have `$JBOSS_HOME`, so that we don't have to have a different script on EAP (where the script would be in `$JBOSS_HOME/bin/standalone.sh`, which in that case is `/opt/eap/bin/standalone.sh`.

```suggestion
exec $JBOSS_HOME/bin/standalone.sh -b 0.0.0.0 $@
```

servers/plain/src/main/docker/Dockerfile
wei-lee: fixed

secondsun:
We need to update a lot of docs too I think now.  By default the build used the in memory database because it was really cheap and quick to verify things.  

Also, perhaps we should make this a build time flag using a profile?

PR #1017, opened 2019-03-14
Fix unifiedpush docker image generation

danielpassos:


danielpassos:
@wei-lee Any thoughts?

PR #1018, opened 2019-05-08
Have APNSCertificate change events use a JMS topic instead of CDI event

danielpassos:
Related issue: [AEROGEAR-9068](https://issues.jboss.org/browse/AEROGEAR-9068)

## How to test?

1. Build the project ignoring the tests
   ```shell
   mvn clean install -DskipTests
   ```
1. Add a new new instance of UniedPush Server in `docker-compose/docker-compose-v2.1.yaml` 
   ```yaml
     unifiedpushserver_2:
    image: aerogear/unifiedpush-wildfly-plain:2.2.2-SNAPSHOT
    volumes:
       - ./helper:/ups-helper:z
    entrypoint: "/ups-helper/exportKeycloakHost.sh"
    depends_on:
       unifiedpushDB:
         condition: service_healthy
    environment:
        POSTGRES_SERVICE_HOST: ${POSTGRES_SERVICE_HOST}
        POSTGRES_USER: ${POSTGRES_USER}
        POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
        POSTGRES_SERVICE_PORT: ${POSTGRES_SERVICE_PORT}
        POSTGRES_DATABASE: ${POSTGRES_DATABASE}
        KEYCLOAK_SERVICE_HOST: ${KEYCLOAK_SERVICE_HOST}
        KEYCLOAK_SERVICE_PORT: ${KEYCLOAK_SERVICE_PORT}
    links:
      - unifiedpushDB:unifiedpush
      - artemis:artemis
      - keycloakServer:keycloak
    ports:
      - 9998:8080      
   ```
1. Run the docker compose. It will run 2 instance of the UPS point to the same postgres and ActiveMQ simulating 2 Openshift pods.
   ```shell
   cd docker-compose && docker-compose -f ./docker-compose-v2.1.yaml up
   ```
1. Create a new project and a iOS variant.
1. Edit cert in the iOS variant (you can change it for the same with the previous one) 
1. See the magics happening between the 2 UPS instance
1. You can also check the ActiveMQ in `http://localhost:8161`. (artemis/simetraehcapa)

### DIFF
--- a/docker-compose/docker-compose-v2.1.yaml
+++ b/docker-compose/docker-compose-v2.1.yaml
@@ -2,7 +2,7 @@ version: '2.1'
 
 services:
   unifiedpushserver:
-    image: docker.io/aerogear/unifiedpush-wildfly:2.1.0
+    image: aerogear/unifiedpush-wildfly-plain:2.2.2-SNAPSHOT
     volumes:
        - ./helper:/ups-helper:z
     entrypoint: "/ups-helper/exportKeycloakHost.sh"
@@ -19,6 +19,7 @@ services:
         KEYCLOAK_SERVICE_PORT: ${KEYCLOAK_SERVICE_PORT}
     links:
       - unifiedpushDB:unifiedpush
+      - artemis:artemis
       - keycloakServer:keycloak
     ports:
       - 9999:8080
@@ -40,3 +41,12 @@ services:
     environment:
       KEYCLOAK_USER: ${KEYCLOAK_USER}
       KEYCLOAK_PASSWORD: ${KEYCLOAK_PASSWORD}
+  artemis:
+    image: vromero/activemq-artemis
+    volumes:
+      - ./artemis-overrides:/var/lib/artemis/etc-override:z
+    ports:
+      - 8161:8161
+      - 61616:61616
+      - 61617:61617
+ 
--- a/model/jpa/pom.xml
+++ b/model/jpa/pom.xml
@@ -82,7 +82,12 @@
                     <artifactId>hibernate-entitymanager</artifactId>
                     <scope>test</scope>
                 </dependency>
-        
+        <dependency>
+    <groupId>javax.xml.bind</groupId>
+    <artifactId>jaxb-api</artifactId>
+    <version>2.2.11</version>
+                        <scope>test</scope>
+</dependency>
                 <dependency>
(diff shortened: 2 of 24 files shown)


docker-compose/docker-compose-v2.1.yaml
```diff
@@ -2,7 +2,7 @@ version: '2.1'
 
 services:
   unifiedpushserver:
-    image: docker.io/aerogear/unifiedpush-wildfly:2.1.0
+    image: aerogear/unifiedpush-wildfly-plain:2.2.2-SNAPSHOT
```
grdryn: Not part of this PR, but I think it would be good to start publishing this to https://quay.io/organization/aerogear instead

danielpassos: Good catch, we can change it before merge, anyway it's using temp SNAPSHOT and will be replaced by the released one!

model/jpa/pom.xml
```diff
@@ -82,7 +82,12 @@
                     <artifactId>hibernate-entitymanager</artifactId>
                     <scope>test</scope>
                 </dependency>
-        
+        <dependency>
```
grdryn: The indentation seems a little off here.

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/cache/SimpleApnsClientCache.java
```diff
@@ -1,13 +1,13 @@
...
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- * 	http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
grdryn: I guess these are html paragraph tags (and probably generated by something), are they? Does the fact that there are 3 of them mean that one is mismatched? :thinking: Or are they all mismatched (should they have closing `</p>` -- I can't really remember html, but I think it's usually obscenely forgiving with that kind of thing)?

secondsun:
@danielpassos I included this pr in https://github.com/aerogear/aerogear-unifiedpush-server/pull/1019

PR #1021, opened 2019-06-12
feat: Universal Container Image

secondsun:
This does a few things.
 * plain and keycloak containers are removed
 * universal container is update
 * README updated with universal container documentation

# Testing
You can run the container with no configuration with this command 
```
docker run -p8080:8080 aerogear/unifiedpush-configurable-container:2.2.2-SNAPSHOT 
```

This should set up a UPS instance at *localhost:8080* with an in-memory database, in vm AMQ, and no security.

To test configuration there are two options test using the docker compose functionality
```
cd docker-compose;
docker-compose -f docker-compose-v2.1.yaml up
```

This should set up a UPS instance at *localhost:9999* with an postgres databse, keycloak, and artemis.  You can view your artemis web app at http://localhost:8161/console/login.  Navigating to UPS at http://localhost:9999 should redirect you to Keycloak.  The KeyCloak login is admin:123 .


danielpassos:
Questions:

1. Should we update the [README getting started](https://github.com/aerogear/aerogear-unifiedpush-server#getting-started)
1. Are we back with h2 (inmemory) DB?
1. Can we change the exposed wildfly port from 9999 to 8080. Myself and @b1zzu had problem with this port before.

secondsun:
@danielpassos  
Answers
1. Yes I'm looking at that today
2. If you don't specify a postgres database, an inmemory h2 database is used.  This is documented here : https://github.com/aerogear/aerogear-unifiedpush-server/blob/bd58beedde06bcb76ce5eb0638e019687bcc7ecb/README.md#container-configuration
3. You can change anything you want by editing the docker-compose.yaml file.  Are you asking if we can make that file use 8080 as the default?

danielpassos:
Yeap, change 9999 to 8080 by default should avoid future problems

PR #1023, opened 2019-06-28
fix: updating postgres driver

secondsun:
Very simple driver update, UPS should run s normal.


databases/prepare_clean_psql_wildfly.sh
```diff
@@ -10,7 +10,7 @@ else
     printf "😂 Awesome, WildFly 11 is configured!\n"
```
danielpassos: Not related of this PR but we should remove any specific version from this kind of file!

PR #1030, opened 2019-08-14
Update the admin ui dependency

danielpassos:


secondsun:
CI will be broken while admin-ui package is being distributed.  I'm going to override and merge.

PR #1034, opened 2019-09-12
Apns push type

f24-gklein:
## Motivation
This PR addresses two aspects (1 feature and 1 bug)
- https://groups.google.com/forum/#!topic/aerogear/CENxfhFVnYk
- https://issues.jboss.org/browse/AEROGEAR-9889

## What
- Add support for apns-push-type field required by ios 13
- Consider ttl for apns pushes

## Why
- apns-push-type is required in apns pushes in ios13. If not set pushes may be delayed or even dropped
- ttl: PushyApnsSender ignores Config#timeToLive, instead it uses a fix TTL of 1 day (default used by Pushy's SimpleApnsPushNotification class). It would be good to consider the ttl provided in org.jboss.aerogear.unifiedpush.message.Config#timeToLive.
BTW: This seems like a regression. In former times this worked correctly before switching from notnoop.apns to pushy. See former solution in UPS 1.1.3 (method: createFutureDateBasedOnTTL): https://github.com/aerogear/aerogear-unifiedpush-server/blob/1.1.3.Final/push/sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/APNsPushNotificationSender.java

## How
- apns-push-type: Upgrade pushy to 0.13.9 which allows specifying PushType as a constructor parameter. Internally it will be sent with the http header
- ttl: use the implementation used back in UPS 1.1.3

## Verification Steps
apns-push-type:
1. Send a push message to an ios13 device with at least sound or alter field set
2. Push is expected to be delivered

ttl:
1. turn of ios device (or flight mode)
2. send push to ios device with a ttl of multiple days
3. wait more than 1 day (and <28 days)
4. turn on device
5.  push must be received

## Checklist:

- [x] Code has been tested locally by PR requester

## Progress
?

## Additional Notes
None

### DIFF
--- a/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
+++ b/push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
@@ -126,12 +126,13 @@ public class PushyApnsSender implements PushNotificationSender {
 
             final String defaultApnsTopic = ApnsUtil.readDefaultTopic(iOSVariant.getCertificate(),
                     iOSVariant.getPassphrase().toCharArray());
+            Date expireDate = createFutureDateBasedOnTTL(pushMessage.getConfig().getTimeToLive());
             logger.debug("sending payload for all tokens for {} to APNs ({})", iOSVariant.getVariantID(), defaultApnsTopic);
 
             tokens.forEach(token -> {
-                final SimpleApnsPushNotification pushNotification = new SimpleApnsPushNotification(token, defaultApnsTopic,
-                        payload, new Date(System.currentTimeMillis() + SimpleApnsPushNotification.DEFAULT_EXPIRATION_PERIOD_MILLIS), 
-                        DeliveryPriority.IMMEDIATE, determinePushType(pushMessage.getMessage()), null, null);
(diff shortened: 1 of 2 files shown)


danielpassos:
Just a heads up on it, I'm doing some final tests (TTL) before we merge it.

push-sender/src/main/java/org/jboss/aerogear/unifiedpush/message/sender/apns/PushyApnsSender.java
```diff
@@ -145,8 +150,33 @@ public void sendPushMessage(final Variant variant, final Collection<String> toke
...
+    	}
+    	return PushType.BACKGROUND;
+	}
+    
+    private boolean isEmpty(final CharSequence cs) {
+        return cs == null || cs.length() == 0;
+    }
 
-    private void handlePushNotificationResponsePerToken(
+	private void handlePushNotificationResponsePerToken(
```
danielpassos: Can you fix the format here?

danielpassos:
@f24-klein Thanks for your contributions. I have tested it in different ways and it works like a charm. Can you fix the format and rebase it before I merge?

danielpassos:
Landed at https://github.com/aerogear/aerogear-unifiedpush-server/commit/fc6cbb09ca1a47dd7886318ba9f52a5f72f6ccc3

PR #1037, opened 2019-09-17
Aerogear 9903

secondsun:
## Motivation
See AEROGEAR_9903, WebPush clients can now create installations in UPS.

## How
We have added the ability for the installation endpoint to receive installations where the device token is a json representation of the web push subscription object generated by the client.

## Verification Steps
This is actually hard and confusing because we don't have a js sdk implemented yet.  I will work with @danielpassos  to create a script that can be run.

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -15,8 +15,6 @@ services:
         POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
         POSTGRES_SERVICE_PORT: ${POSTGRES_SERVICE_PORT}
         POSTGRES_DATABASE: ${POSTGRES_DATABASE}
-        KEYCLOAK_SERVICE_HOST: ${KEYCLOAK_SERVICE_HOST}
-        KEYCLOAK_SERVICE_PORT: ${KEYCLOAK_SERVICE_PORT}
         ARTEMIS_USER: ${ARTEMIS_USER}
         ARTEMIS_PASSWORD: ${ARTEMIS_PASSWORD}
         ARTEMIS_SERVICE_HOST: ${ARTEMIS_SERVICE_HOST}
@@ -26,7 +24,6 @@ services:
     links:
       - unifiedpushDB:unifiedpush
       - artemis:artemis
-      - keycloakServer:keycloak
     ports:
       - 9999:8080
   unifiedpushDB:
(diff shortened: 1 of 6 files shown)


docker-compose/docker-compose.yaml
```diff
@@ -15,8 +15,6 @@ services:
         POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
         POSTGRES_SERVICE_PORT: ${POSTGRES_SERVICE_PORT}
         POSTGRES_DATABASE: ${POSTGRES_DATABASE}
-        KEYCLOAK_SERVICE_HOST: ${KEYCLOAK_SERVICE_HOST}
```
danielpassos: Why it was removed?

model/api/src/main/java/org/jboss/aerogear/unifiedpush/utils/KeyUtils.java
```diff
@@ -42,12 +48,29 @@ public static PrivateKey loadPrivateKey(String privateKey) throws  NoSuchAlgorit
     public static PublicKey loadPublicKey(String publicKey) throws NoSuchAlgorithmException, InvalidKeySpecException {
         byte[] decodedPublicKey = Base64Encoder.decode(publicKey);
         KeyFactory keyFactory = KeyFactory.getInstance(ALGORITHM, PROVIDER);
-        ECParameterSpec parameterSpec = ECNamedCurveTable.getParameterSpec(CURVE);
+        ECParameterSpec parameterSpec;
+        parameterSpec = ECNamedCurveTable.getParameterSpec(CURVE);
```
ziccardi: Why splitting this to two rows?

docker-compose/docker-compose.yaml
secondsun: By accident for testing stuff.

model/api/src/main/java/org/jboss/aerogear/unifiedpush/utils/KeyUtils.java
secondsun: I think my intellij has drunkenly changed its autoformat rules.

docker-compose/docker-compose.yaml
```diff
@@ -26,7 +26,7 @@ services:
     links:
       - unifiedpushDB:unifiedpush
       - artemis:artemis
-      - keycloakServer:keycloak
+    - keycloakServer:keycloak
```
danielpassos: 🧐

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/WebPushRegistration.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return the endpoint
+     */
+    public String getEndpoint() {
+        return endpoint;
+    }
+    /**
+     * @return the keys
+     */
+    public PushKey getKeys() {
+        return keys;
```
ziccardi: This will return a mutable internals. If we can't make it immutable, we should return a copy

PR #1044, opened 2019-10-17
Aerogear 8829 8831 apns token

secondsun:
## Motivation
Add support for APNS token signed push messages

## How
 * A second iOS variant type was created, iOSTokenVariant.
 * A second pushy sender was created
 * A common "IsAPNSVariant" interface was created for general APNS connectivity
 * JMS queues are unaffected
 * Tests were written as necessary 

## Verification Steps
 0.  Setup a APNS token project in Apple's dev console
 1. Create a variant using the RESTful API
 2. Verify that it appears in the admin-ui console; no work was done, but this shows it is ther
 3. Configure and register a mobile application using this variant
 4. Use the push sender API to send a message


model/api/pom.xml
```diff
@@ -53,6 +53,12 @@
         <groupId>com.fasterxml.jackson.core</groupId>
         <artifactId>jackson-annotations</artifactId>
       </dependency>
+        <dependency>
```
danielpassos: Why we need this in the model?

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/IsAPNSVariant.java
```diff
@@ -0,0 +1,15 @@
+package org.jboss.aerogear.unifiedpush.api;
+
+import java.io.Serializable;
+
+/**
+ * There are two APNS Variants, iOSVariant and iOSTokenVariant.  However, they also have some metadata that is shared
+ * as well as needing to have their updates observed by SimpleApnsClientCache.  This is a interface that defines the
+ * common fields to update the cache as well as mark a variant as having APNS information.
+ */
+public interface IsAPNSVariant extends Serializable {
```
danielpassos: I like the idea but the name of the interface is so confused looking into the code. Maybe we should just use `IAPNSVariant`, `APNSVariantType`, `APNSVariantInterface` or just `APNSVariant`.

@ziccardi Any suggestion for a better name?

ziccardi: It is an interface implemented by all the iOS Variants, so why not simply `iOSVariantInterface` ?
Moreover: the `production` member variable and its setter and getter are duplicated into `iOSVariant` and `iOSTokenVariant`. The same for the `getType` method. I would suggest to replace the interface (or use together) with an abstract base class:

```suggestion
public abstract class AbstractIOSVariant implements Serializable {
   private final VariantType type;
   private boolean production;
   
   protected AbstractIOSVariant(final VariantType type) {
      this.type = type;
   }
   
   public getType() {
      return this.type;
   }
   
   public boolean isProduction() {
        return production;
   }

   public void setProduction(boolean production) {
        this.production = production;
   }
}
```

model/api/pom.xml
secondsun: It has utility classes for validating the variant.

model/api/src/main/java/org/jboss/aerogear/unifiedpush/api/IsAPNSVariant.java
secondsun: Seems reasonable; the windows variants did the same thing.

danielpassos:
# UnifedPush Server

| Variant created  | Device registered  | Sending message  |
|---|---|---|
| Screen Shot 2019-10-28 at 16 50 16 | Screen Shot 2019-10-28 at 17 13 40 | Screen Shot 2019-10-28 at 17 13 56
 |

# Cordova iOS app

| Device registered | Message received  |
|---|---|
| Device registered | Message received |

PR #1053, opened 2019-12-11
initial docs config

secondsun:
## Motivation
See AEROGEAR-10101

This PR creates UI config points and exposes a json configuration for the UI to consume.  

## What
* Two environment variables have been added, DOCS_PATH which is a path to the doc-links.json file and UPS_DISABLED which is a comma separated list of variants to not allow.
* A default doc-links.json is embedded into the source code so that default values are provided for doc links. If any key misses from the provided `doc-links.json`, the default one is used.

## Verification Steps
Add the steps required to check this change. Following an example.
 
1. `mvn clean install`
2. configure a doc-links.json file (there is a default file in the PR)
3. Set the environment variable to point to the path of your file
4. Run UPS
5. Navigate to http://localhost:8080/rest/ui/config and confirm the JSON matches your values

If you are having trouble, I would suggest copying the doc-links.json file to your tmp directory and editing that.  THen you can mount your tmp directory in a docker container and load that file using the environment variable.

Ex
```
docker run --rm -p 8080:8080 -it  -v /tmp:/tmp -e DOCS_PATH=/tmp/doc-links.json -e UPS_DISABLED=ios,android,web_push aerogear/unifiedpush-configurable-container:2.5.1-SNAPSHOT
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -114,6 +114,7 @@ The Unified Push Server can be configured with either System Properties (passed 
 | `ARTEMIS_PASSWORD` |  Password for AMQP server.
 | `ARTEMIS_USERNAME` |  Username for AMQP server.
 | `UPS_DISABLED` | a comma separated list of variants to be disabled per their variant type.  See `org.jboss.aerogear.unifiedpush.rest.annotations.DisabledByEnvironment` for more details
+| `DOCS_PATH` |  File system path for documentation links json file. See [docs-links](./jaxrs/src/main/resources/doc-links.json) as an example
 |===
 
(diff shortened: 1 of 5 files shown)


secondsun:
You may need https://github.com/aerogear/unifiedpush-admin-ui/pull/45 to be installed locally before this PR will build

secondsun:
The build fails because we are using a snapshot of the UI, once the UI PR is approved and merged I will make a release and update this PR

Boomatang:
Changes looks good the defaults load nicely and the docs link file can be hot swapped with new values.

PR #1059, opened 2020-01-17
Aerogear 10135

ziccardi:
## Motivation
Added Swift SDK docs


website/docs/configuring_clients.md
```diff
@@ -51,6 +51,86 @@ Now, be sure you click in `Remote notifications` under `Background Modes` to ena
...
+>
+>target 'Demo App' do
+>  # Comment the next line if you don't want to use dynamic frameworks
+>  use_frameworks!
+>
+>  # Pods for Demo App
+>
+>  target 'Demo AppTests' do
+>    inherit! :search_paths
+>    pod 'AeroGearPush-Swift', '~> 3.1.0'
```
Aindriunannerb: Hi Massimo,
Here you specify '3.1.0',  but above under 'Uncomment the platform as suggested' you specify '3.0.0'.
which one is correct?

website/docs/configuring_clients.md
```diff
@@ -8,8 +8,155 @@ title: Configuring Clients
...
+Now, be sure you click in `Remote notifications` under `Background Modes` to enable it:
+
+>![CreateDemoApp](./assets/ios/checkBackgroundModes.png)
+
+### Registering your app
+
+Before you can receive any _push notification_ the first thing needed is to register the app to the _Unified Push Server_.
+To be able to perform such operation, you will need the following information:
+  * The URL of the _Unified Push Server_ instance
+  * The ID of an iOS variant you previously created in UPS (to see how, look [here](./configuring_variants#ios))
```
Aindriunannerb: Hi Massimo, 
The link "here" does results in a 404 page not found message.

website/docs/configuring_clients.md
ziccardi: 3.1.0 is the correct one. However it is not published yet (I'm waiting to have access to the pod).
That mean that, when you will test it, you will have to link the pod locally.

Good catch!

website/docs/configuring_clients.md
ziccardi: Just tried it and it works.
How are you looking at the docs? Are you running the webiste locally with `npm run start` ?

website/docs/configuring_clients.md
```diff
@@ -8,10 +8,342 @@ title: Configuring Clients
...
+
+2. Do the same for the `SceneDelegate.swift`
+
+```swift
+@available(iOS 13.0, *)
+class SceneDelegate: UIResponder, UIWindowSceneDelegate {
+...
+```
+
+3. In the `AppDelegate` mark with the `@available` keyword each method referring to the `UIScene*` classes
```
secondsun: I think this line sounds more natural as : 

 In the `AppDelegate` mark each method referring to the `UIScene*` classes with the `@available` keyword.

ziccardi:
> I worked through these changes and setup a second app this time registering by 'pushconfig.plist'. I got it registering and receiving push notifications no problem.
> 
> however I included ' import AeroGearPush' at the top of my AppDelegate.swift file, I didn't see an instruction to do this in your docs. Is this a required import?

Yes, good catch!

> 
> when creating the 'pushconfig.plist' file I was unsure of which 'types' to use for the url, variantID and variantSecret so I just guessed string and it worked, would it be necessary to specify this in the docs?

The docs says "The plist file must contain three **string** properties"

PR #1067, opened 2020-02-07
Forcing chameleon to use https during builds

secondsun:
## Motivation
Maven has removed support for http based urls, some of our testing libraries are still using them; this overrides those libraries.

## What
added https overrides for maven central in the root pom.

## Why
This was done in the root pom instead of aerogear-parent because we would like to move away from aerogear-parent

## Verification Steps
Delete your local maven cache (~/.m2/repository) and build the project.  It should build without errors/test failures.

## Checklist:

- [ x ] Code has been tested locally by PR requester

## Additional Notes

If you get an error about docker, this means that the code has build succesfully and you can't build the docker image.  This technically means that the code has worked.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" ?>
 <!--
   JBoss, Home of Professional Open Source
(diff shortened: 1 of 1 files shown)


secondsun:
I'm closing without merging; it doesn't seem like this pr fixes the issue I thought it did.

PR #1076, opened 2020-02-19
Added frontpage, getting help and fixed the footer

ziccardi:
## Motivation
Added a frontpage so that we can start thinking on replacing the current website

PR #1078, opened 2020-02-20
Frontpage V2

ziccardi:
## Motivation
Created a frontpage for the new website

An alternative proposal to #1076, more appealing and modern


Aindriunannerb:
Looks very nice, a higher resolution logo to replace the one in the bottom left of the page would be nice. It would be nice to also keep that info diagram from the first version of the frontpage, But I'm not sure how/where you would implement it.

website/website/pages/en/index.js
```diff
@@ -92,117 +28,60 @@ class Index extends React.Component {
...
-
-    const TryOut = () => (
-      <Block id="try">
+    const Features = () => (
+        <div>
+      <Block layout="threeColumn">
+          {[
+              {
+                  content: 'A server written in Java which allow to send push notifications to different mobile operating systems and web application',
+                  image: `${baseUrl}img/notification.svg`,
```
Aindriunannerb: possible to make the image also a link? This is what I clicked first. If its a pain, would making them into cards be a better option?

ziccardi: Done!

ziccardi:
@Aindriunannerb I removed the icon from the footer since it didn't add any value to the website. The diagram is still in the repo, we will probably add it back in another documentation page, maybe an intro one.
I should have fixed all the issues reported: mind giving another look?

website/website/pages/en/index.js
```diff
@@ -47,30 +47,35 @@ class Index extends React.Component {
             content: 'A Java library that allows the integration of the UnifiedPush Server into your native Android application',
             image: `${baseUrl}img/java.svg`,
             imageAlign: 'top',
+            imageLink: '/docs/configuring_clients#android',
```
Aindriunannerb: Awesome stuff, much more smoother user experience!

website/docs/introduction/how-it-works.md
```diff
@@ -0,0 +1,38 @@
+---
+id: how-it-works
+title: How the UnifiedPush Server works
+---
+
+The _AeroGear UnifiedPush Server_ can be seen as a broker that distributes push messages to different 3rd party Push Networks. The graphic below gives a little overview:
+
+![Load Firebase Assistant](/img/ups-overview.png)
```
Aindriunannerb: is it possible to make the image a little big bigger?, the text is small and the light blue colour on the white background makes it a little bit hard to read.

ziccardi: Don't think so... It already occupies the whole available width

PR #1107, opened 2020-06-10
Rn doc update

secondsun:
minor updates to rn docs

### DIFF
--- a/website/docs/clients/reactnative.md
+++ b/website/docs/clients/reactnative.md
@@ -17,7 +17,7 @@ Once we've done this we can begin configuring our native projects.
 
 ## Android Native Project configuration
 
-Open the `android` project found in your React Native project root in Android Studio, and begin integrating Firebase Cloud messaging into the application. This is most easily done using the Firebase Assistant, and you can find that documentation in the [Android client documentation](./android-client)
+Open the `android` project found in your React Native project root in Android Studio, and begin integrating Firebase Cloud Messaging into the application. This is most easily done using the Firebase Assistant, and you can find that documentation in the [Android client documentation](./android-client)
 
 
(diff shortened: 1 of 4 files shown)


website/docs/clients/reactnative.md
```diff
@@ -160,23 +160,26 @@ We need to call `ups.init` to initialize our library. This method takes one para
...
 A message handler is a function that takes a single parameter, the message. This message is your push notification data. You can register the message handler with the following code : 
 
-```
+```javascript
 const ups = new RNUnifiedPush();
 ups.registerMessageHandler(message => console.log("You have received a push message.",  JSON.stringify(message)));
 ```
 
+You will see different push objects when you run your application on Android vs when you run it on iOS. Message handlers can switch on the `aps` property of the message. This value is set on iOS and is not set on Android. 
+
```
ziccardi, line 182: I believe that instead of switch on the `aps` property, would be better to use this: https://reactnative.dev/docs/platform-specific-code

WDYT?

website/website/package.json
```diff
@@ -13,5 +13,8 @@
   },
   "dependencies": {
     "remarkable-admonitions": "^0.2.2"
+  },
+  "resolutions": {
+    "babel-plugin-dynamic-import-node": "2.3.0"
```
ziccardi, line 18: Do we need this?

secondsun: Yes. There is a bug in some combinations of docusaurus and babel right now. I can find the thread I got this from if you want to read up on it.

secondsun: https://github.com/facebook/docusaurus/issues/2637

secondsun:
@ziccardi updated, care to take another look?

PR #1110, opened 2020-06-23
fix!: replacing error string in responses with json objects

secondsun:
@ziccardi I've started by replacing all of the errors that just returned a string. Would you like to take some time to discuss how to beef up the `UnifiedPushError` object?

### DIFF
--- a/docker-compose/docker-compose.yaml
+++ b/docker-compose/docker-compose.yaml
@@ -2,7 +2,7 @@ version: '2.1'
 
 services:
   unifiedpushserver:
-    image: quay.io/aerogear/unifiedpush-configurable-container:master
+    image: aerogear/unifiedpush-configurable-container:2.5.1-SNAPSHOT
     volumes:
        - ./helper:/ups-helper:z
     entrypoint: "/ups-helper/exportKeycloakHost.sh"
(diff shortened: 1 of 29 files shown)


secondsun:
@ziccardi I added more fixes. I think it is ready for a proper review

ziccardi:
# Verification

## Applications:
- [ ] list
- [ ] found
- [ ] not found
- [ ] create
- [ ] update complete (all fields)
- [ ] update partial (some field)
- [ ] delete application
## Variants
- [ ] list
- [ ] find with wrong appId with correct variant id and type
- [ ] find with good appId with wrong variant id
- [ ] find with good appId with wrong variant type
- [ ] create android
- [ ] create ios
- [ ] create ios_token
- [ ] create webpush
- [ ] update complete and partial android
- [ ] update complete and partial ios
- [ ] update complete and partial ios_token
- [ ] update complete and partial webpush
- [ ] delete android
- [ ] delete ios
- [ ] delete ios_token
- [ ] delete webpush

secondsun:
@ziccardi So the iOS developer field bit is interesting; it doesn't use an iOS variant object but an iOS application form which doesn't have the developer property. So that is the correct behavior.

PR #1115, opened 2020-07-10
build: wildfly update

secondsun:
## Motivation
I noticed that we're a couple of version behind on the wildfly base image, so I've bumped it up.

## Why
General update hygene.

## Verification Steps
Ok, this is going to be important that we check our bases so I would like for all of the following this be the testing procedure
 
1. Create an Application and one of each type of variant 
2. Use the cookbook to verify each type of variant can send and receive push messages
3. Verify WebPush, Android, iOS tokens, and iOS Certificates

## Checklist:

- [ ] Verified With using enmasse in k8s/openshift
- [ ] Verified With Keycloak security enabled
- [ ] Verified With Artemis AMQ based JMS
- [ ] Verified With internal JMS and no Keycloak

## Additional Notes

This is a small change work wise but a big change implication wise. This PR will stay open a while, but everyone is welcome to take it on in small chunks and report your progress below.


ziccardi (approved):
Fix conventional commit. Other than that, LGTM

